PR #77, opened 2018-01-10
Allow admins to delete public things from other admins

JulianKniephoff:
Closes #30. As the comment in pull request #78 suggests, this one pulls in that one. :wink: 

Every admin (= Opencast admin or user with annotate-admin action on the video) can now delete anything and everything that is public in the "structured annotation panel".

PR #80, opened 2018-01-11
Removed last vestiges of "annotations-tool" terminology

JulianKniephoff:
I took the freedom to rename all the remaining things called `annotations-tool` (and variations of that) to `annotation-tool` (and the corresponding variation of that ...). For the lazy readers (which includes me): The difference between the two is the plural s.

This PR does nothing else but I would still like to have a second pair of eyes on it, because it is so big.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,13 @@ The software devides currently into two parts:
 
 * The frontend component, containing the annotation webapp.
 * The Opencast Backend modules, that need to be integrated into an Opencast installation. A recent version of the i
-  frontend is already included in the entwine-annotations-tool module. 
+  frontend is already included in the annotation-tool module.
 
 ## [Installation](documentation/opencast-installation.md)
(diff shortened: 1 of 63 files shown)


JulianKniephoff:
Should be fixed. :)

rrolf:
This still (or again? as I just merged pr77) has merge conflicts.

JulianKniephoff:
Again. ;) #77 and #81 both introduced changes involving the global `annotationsTool` object which this PR renames (among other things), so it needs to be updated for both of them (or the other way around, depending on the merge order). I would say we merge before this one now, too, so this PR only needs to be updated once.

I will leave the merge conflicts in this one until then to remind me; we are lucky that there are some, it might as well just have silently failed. :wink:

JulianKniephoff:
Fixed the merge conflicts and also changed all the newly introduced references to `annotationsTool` by #77 and #81.

PR #92, opened 2018-01-29
Enable use of the annotation tool as Karaf feature

JulianKniephoff:
See the documentation (included in this PR) for how to use it. :wink:

### DIFF
--- a/documentation/build-frontend.md
+++ b/documentation/build-frontend.md
@@ -8,7 +8,7 @@ At first you need to checkout the annotation tool with GIT:
 
     mkdir /opt/annotation
     cd /opt/annotation
-    git clone https://github.com/JulianKniephoff/annotation-tool.git
+    git clone https://github.com/opencast/annotation-tool.git
 
 Make sure you have installed [Node.js](https://nodejs.org/en/download/), which you might find as a binary package in a repo also, and [Grunt](http://gruntjs.com/getting-started) on your machine.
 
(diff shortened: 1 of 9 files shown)


JulianKniephoff:
Re: yesterdays discussion between @rrolf, @lkiesow and me about "Wouldn't `feature:install` be the preferred installation method?":

Maybe, but:

* As I suspected, `filter:install` does not survive a restart. This is, as @lkiesow already suspected, because Opencast explicitly configures Karaf to delete its bundle cache on every restart. See `karaf.clean.cache` in `etc/system.properties`).
* It is broken right now. x) After the `filter:install` call, everything is broken horribly<sup>[1](#breakage)</sup>. Restarting Opencast after that, with `karaf.clean.cache = false`, fixes everything, and the annotation tool is apparently installed successfully, but this is of course suboptimal and in light of the first point also "impossible" (presuming that the reason @lkiesow gave for clearing the cache is still valid and is not addressed otherwise).

<sup><a name="breakage">1</a></sup>: First I get dozens of exceptions about inserting rows into `mh_bundleinfo`; apparently some generated bundle names are too long for that table. I'm investigating.

After allowing for longer strings in the corresponding column, the `feature:install` command itself succeeds, but visiting both, the admin UI and the annotation tool in the browser shows a 500 error page with a long Java exception stack trace about a closed persistence environment. I am investigating this, too.

JulianKniephoff:
Since I can't reproduce my `feature:install` problems, @rrolf's build problems are "resolved" (by noting that they are problems with the Opencast Nexus, see [this post on the dev mailing list](https://groups.google.com/a/opencast.org/forum/#!topic/dev/KyWEmPgd9yE)) and the discussion about `feature:install` vs. boot feature configuration is settled for now, this PR should be good to go, right?

I also changed the documentation to reflect all we have learned now, incorporating @rrolf's suggestions and corrections.

Issue #74, opened 2017-12-11
German translation missing for loop controller


JulianKniephoff:
Originally reported by: **Thomas Schraml (Bitbucket: [TSchraml](https://bitbucket.org/TSchraml), GitHub: [TSchraml](https://github.com/TSchraml))**

----------------------------------------

The loop controller and the displayed loops in the track view are still in English, whereas the rest of the tool is translated to German.

![Bildschirmfoto 2017-12-11 um 12.49.49.png](https://bitbucket.org/repo/7kKKrp/images/3195951454-Bildschirmfoto%202017-12-11%20um%2012.49.49.png)

----------------------------------------
- Bitbucket: https://bitbucket.org/opencast-community/annotation-tool/issue/74

JulianKniephoff:
The translations are actually already there, they are just not loaded, yet, when the strange plugin mechanism that is used for the loop controller loads this component. :thinking:

Issue #42, opened 2017-06-29
Loop controller: keyboard does not activate it correctly


JulianKniephoff:
Originally reported by: **Rüdiger Rolf (Bitbucket: [rrolf](https://bitbucket.org/rrolf), GitHub: [rrolf](https://github.com/rrolf))**

----------------------------------------

Make sure the loop controler checkbox is focused for the keyboard. Use the spacebar to activae the checkbox: the checkbox is checked, but the loop-controller settings do not show up. Probably this is triggered with a "clicked" and not with "checked" event.

----------------------------------------
- Bitbucket: https://bitbucket.org/opencast-community/annotation-tool/issue/42

Issue #24, opened 2017-03-21
UI Templates


JulianKniephoff:
Originally reported by: **Rüdiger Rolf (Bitbucket: [rrolf](https://bitbucket.org/rrolf), GitHub: [rrolf](https://github.com/rrolf))**

----------------------------------------

There should be UI templates to switch between different work modes:

* Fullscreen of Timeline
* Switching the timeline and the new annotation area

----------------------------------------
- Bitbucket: https://bitbucket.org/opencast-community/annotation-tool/issue/24

JulianKniephoff:
In combination with #27, we could even make the list of UI templates extensible by the user. ;) #randomthoughts

PR #94, opened 2018-01-31
Refactored and improved the loop controller

JulianKniephoff:
Despite what I said in #74, this turned out not to be necessary to fix the missing translations ... I really just overlooked them ... :sweat_smile: 

But it also doesn't hurt, especially in view of #24.

As an "apology" I threw in a fix for #42, which was easy having looked at the Loop Controller a bit closer now. :wink:

### DIFF
--- a/documentation/opencast-installation.md
+++ b/documentation/opencast-installation.md
@@ -41,7 +41,7 @@ where `${version}` is to be replaced by the version of the annotation tool.
 
 You can conveniently install it using the following Karaf shell commands:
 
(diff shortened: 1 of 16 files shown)


rrolf:
Could you create the loop view as a pinned row that does not move if I scroll the tracks?

JulianKniephoff:
Phew, I mean probably I could, but that would mean mucking around with the DOM and/or CSS a lot, I think. I could make some experiments, but I would prefer you make another issue for that, since that is definitely outside the scope of this PR, imo. :slightly_smiling_face:

rrolf:
Okay. I created two new issues for this.

PR #106, opened 2018-03-14
Some major refactorings

JulianKniephoff:
Normally I would just push my `nitpicks` branch at some point because it should only contain trivial changes, but somehow this spiraled out of control this time and includes some large refactorings that came out of trying to get the player into GoldenLayout.

This PR should not change the behavior of the tool fundamentally, but the changes are significant enough so that I want a second pair of eyes looking out for any obvious breakage I might have caused. :wink:


rrolf:
I cannot edit public categories anymore as an admin

rrolf:
I cannot change the color of an category anymore

JulianKniephoff:
@rrolf can you confirm that the second problem only happens once? That is, after having changed the color once unsuccessfully once, trying it again actually works?

And also that even after the first try, a refresh shows that it actually **did** work in the backend, so to speak; it just wasn't correctly updated in-place?

JulianKniephoff:
Alright, both issues should be fixed now.

PR #109, opened 2018-03-20
Refactor video loading, 2nd try

JulianKniephoff:
More preparatory refactorings for the layout management code ...

**NOTE**: This subsumes #106, so that PR should be reviewed and merged first.


JulianKniephoff:
Note that with this PR, the video in the "`grunt dev`-verion" is no longer hard coded to be at `resources/aav1.mp4`. You can (and have to) give a URL to a video with the `video` URL parameter, like so:

    http://localhost:9001?video=resources/aav1.mp4

This would restore the old behavior, assuming there actually is a file under that path in the `www` directory. Absolute (and thus also foreign origin) URLs should work fine, too.

JulianKniephoff:
Player control issue: #112.

Issue #27, opened 2017-03-21
Store UI settings in the local storage of the browser


JulianKniephoff:
Originally reported by: **Rüdiger Rolf (Bitbucket: [rrolf](https://bitbucket.org/rrolf), GitHub: [rrolf](https://github.com/rrolf))**

----------------------------------------

general UI settings (like which UI elements is visible and not) and video specific settings (like which annotations are currently open) should be stored in the local storage so that when a user comes back to a recording it is more or less in the same state where he left it.

----------------------------------------
- Bitbucket: https://bitbucket.org/opencast-community/annotation-tool/issue/27

JulianKniephoff:
The layout is restored, but many other things are not. Examples:

* Player position
* Track selection/visibility
* Which annotations are (not) expanded
* etc.

I guess we should specify what parts of the UI state are most important to the "stakeholders" with respect to the 2.0 release, and create dedicated issues for these, but I would like to keep this general issue as a reminder; in my dream version of the tool, the whole UI state would be persisted, but that will very probably not happen for 2.0. :wink:

JulianKniephoff:
@ebbertd can you briefly specify what things you want to be persisted in particular for you to consider this issue resolved in the context of the last "sprint" we discussed?

The list I have in my head currently contains the following items:

* [ ] Annotation visibility (i.e. which categories are shown/hidden, including whether or not to show free text annotations)
* [ ] Track visibility (i.e. which tracks has the user selected to see in the track manager)

These are the obvious things in my mind; I mentioned a few other things above, but I would like to hear from you specifically what you consider "in scope" at the moment. :wink:

ebbertd:
@JulianKniephoff as the UI layout is already persisted the following items are left:

- [ ] Annotation visibility as described by you
- [ ] Track visibility as described by you
- [ ] Collapsed / expanded annotation (which annotations in the list are expanded and which are collapsed)

dagraf:
@JulianKniephoff and @ebbertd if I may suggest additional things:
- pause video during writing (i.e. activated or not)
- automatically expand activated annotations (i.e. activated or not)

PR #113, opened 2018-03-26
GoldenLayout

### LINKED ISSUE #23: Better use available UI space
Originally reported by: **Rüdiger Rolf (Bitbucket: [rrolf](https://bitbucket.org/rrolf), GitHub: [rrolf](https://github.com/rrolf))**

----------------------------------------

If UI elements are hidden, the available space is not used. 

The remaining UI elements should make better use of the free UI space

----------------------------------------
- Bitbucket: https://bitbucket.org/opencast-community/annotation-tool/issue/23

### LINKED ISSUE #42: Loop controller: keyboard does not activate it correctly

### LINKED ISSUE #90: Default layout configuration is not honored on initial page load
No matter what `annotation-tool-configuration` says about the views visible by default, the main layout is not adapted.

Specifically, when both the list and annotate view are hidden by default, the video and the timeline still only take up the left side of the screen.

Enabling one of those views and then hiding it again, fixes the problem.

### LINKED ISSUE #93: Timeline too short when hiding the loop controller
When enabling the loop controller, the timeline shrinks to accommodate it. When you hide it again after that, the timeline remains (too) small.

### LINKED ISSUE #123: Annotations don't open automatically anymore
Somehow the feature requested in #15 doesn't work anymore.

### LINKED ISSUE #124: Deleting categories behaves strangely in combination with pagination
When you have lots of categories, the structured annotation panel paginates them. When you then delete a category on a page that is not the last, the categories on the page after it come sliding onto the current page and are displayed in a second row of annotations.

JulianKniephoff:
Finally integrate everything halfway decently with GoldenLayout!

#### Must have

* [x] The list view must be scrollable when it is longer than its space
* [x] The layout container is too large
* [x] The titles of the tabs could be more friendly and should be translated
* [x] Hide the loop controller by default again
* [x] Restore the ability to show/hide views
  * [x] Make the menu entries work again
  * [x] Don't leak memory or lose/corrupt state when closing/reopening views

#### Nice to have

* ~~Can we put the "window headers" in the tabs?~~
  See #130
* ~~Split the annotate view into two (free text vs. structured)~~
  See #129
* [x] Make the category tabs use their entire space
* [x] Make the timeline look less awkward without a lot of tracks
* [x] Clean up some more of the old layout management code
* ~~Make the default layout more sensible~~

#### Out of scope

Mainly because they have their own issue already

* Save the changes the user made to the layout<br>This is basically #27
* Provide (more than one) preconfigured layouts<br>This is #24 

### Affected issues

This PR will probably fix/close/invalidate a few issues and raise some new ones.

* Obviously it closes #23
* Interestingly enough, it apparently also fixes #42
* Closes #93
* Will soon fix #123
* Fixes #90
* I think it might also fix #124


JulianKniephoff:
Outstanding tasks after first review round:

* [ ] The background of the active track in the timeline is too short when the timeline is very long
* [ ] Close views on clicking the corresponding menu entry
* [ ] Reset views
* [ ] Add views on click on the corresponding menu entry in some default position
  * [ ] Maybe add a small note to the menu that prompts the user to click or drag
* [ ] Turn off the "pop out" mechanism for now

JulianKniephoff:
Thinking about it a bit more, the point "reset views" is basically the same as #24, right? One of the UI templates could just be "Default". (Or we could think of a more appropriate name for this. ;D)

With this in mind I would say that this, too, is out of the scope of this PR.

@rrolf you can now take another look. :smile:

ebbertd:
@JulianKniephoff Having the current view as the default would be fine.

PR #118, opened 2018-03-27
Fix list scrolling

JulianKniephoff:


JulianKniephoff:
@rrolf the "list view does not scroll back" issue should be fixed, can you confirm that?

And how exactly can I reproduce the other issue? I seem to be unable to do so. :thinking:

PR #150, opened 2018-05-29
Simplify view initialization some more

JulianKniephoff:
Some preparatory changes necessary for #27 and maybe #24.


JulianKniephoff:
This is now contained in #152.

Issue #151, opened 2018-06-18
Make the list of UI templates configurable


JulianKniephoff:
At keast in some kind of configuration file, but preferably by the user in the UI.

Note that this idea originally came out of #24, but due to the architecture of the application, this---again---more complicated than it should be. Thus this dedicated issue.

PR #152, opened 2018-06-18
UI Templates

### LINKED ISSUE #24: UI Templates

JulianKniephoff:
This implements the requested UI templates. The system is also semi-extensible, but there are a few steps involved. See #151 for the full customizability.

This contains and supercedes #150 and closes #24.

### DIFF
--- a/frontend/build/profiles/integration/annotation-tool-configuration.js
+++ b/frontend/build/profiles/integration/annotation-tool-configuration.js
@@ -119,20 +119,6 @@ define(["jquery",
          */
         var Configuration = {
             /**
-             * List of possible layout configuration
-             * @memberOf module:annotation-tool-configuration.Configuration
-             * @type {Object}
-             */
-            LAYOUT_CONFIGURATION: {
-                /** default configuration */
-                DEFAULT: {
-                    list     : true,
-                    annotate : true,
-                    loop     : false
-                }
-            },
-
-            /**
              * The minmal duration used for annotation representation on timeline
              * @alias module:annotation-tool-configuration.Configuration.MINIMAL_DURATION
              * @memberOf module:annotation-tool-configuration.Configuration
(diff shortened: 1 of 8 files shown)


rrolf:
I guess we should discuss what useful templates would be. The "alternative" Template looks more like a proof of concept than a reasonable setting. 

"Templates" should be named "Layout"

I guess that templates should be set based on usage scenarios:
- "Standard" could become "Annotating (default)"
- "Reviewing" Timeline fills the lower half. Video to the top left Annotation list to the top right.
- "Annotating with Loops" is "Annotating (default)" + the loop controler window somewhere very small
- "Fullscreen Timeline" can stay as it is.

A "custom" layout that will be stored in the browsers local storage is probably a follow up ticket?

JulianKniephoff:
I made some more changes:

* Changed the title of the menu to "Layouts"
  * I used "Layouts" in the German translation as well, since there is IMO no good idiomatic translation and layout should be a well established term of art by now; I'm of course open for alternative suggestions, though.
* I added the layouts @rrolf suggested
  * I kept "Alternative", though, since that was explicitly mentioned in the original issue. Granted, the name is less than ideal, but we might be able to come up with an alternative. If this layout is really not wanted (anymore) I can of course remove it, still.

The "Custom" layout is being implemented as part of my work on #27. It does not belong in this PR I think. I will make a PR about that part independent of the rest of that issue, once it is done.

PR #154, opened 2018-07-11
Fix artifactPrefix for 5.0 and newer

mtneug:
The `artifactPrefix` assumes that matterhorn is simply renamed to "opencast"; instead "matterhorn-" is just dropped form the `artefactId`. This also adds the `opencast.osgi.jndi.service.name` property and updates the documentation accordingly. By default the main `pom.xml` file is now configured for Opencast 5.0.

### DIFF
--- a/documentation/opencast-installation.md
+++ b/documentation/opencast-installation.md
@@ -30,9 +30,9 @@ In this manual we use `<annotationtool-dir>` for the base dir of the Annotation 
 This should build the frontend, include it into the Opencast modules and copies the JARs
 to your Opencast installation.
 
-Note that if you build with `opencast.version` set to something `>= 5`, you also have to specify
-`-Dopencast.artifactPrefix=opencast` because of the transition away from the `matterhorn` name
-between versions `4` and `5`.
+Note that if you build with `opencast.version` set to something `<= 4`, you also have to specify
+`-Dopencast.artifactPrefix="matterhorn-" -Dopencast.osgi.jndi.service.name=jdbc/matterhorn` because
+of the transition away from the `matterhorn` name between versions `4` and `5`.
 
 #### As a Karaf Feature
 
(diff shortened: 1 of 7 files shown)


JulianKniephoff:
Are you sure this is right? I was using it successfully with the `opencast-` prefix this whole time and building your branch kind of supports this:

```
[ERROR] Failed to execute goal on project opencast-annotation-api: Could not resolve dependencies for project org.opencast:opencast-annotation-api:bundle:1.1-SNAPSHOT: Could not find artifact org.opencastproject:common:jar:5-SNAPSHOT
```

But now that you brought this topic up again: The way I did this back then is an ugly hack anyway and I would like to wrap this at least into build profiles, at least. I might just as well just do this now. ^_^

The "proper" solution would be to have an Ant task or some kind of script set these properties automatically based on the resolved Opencast version, but I can't find a nice way to do this right now. :thinking:

mtneug:
OK, I am confused now 😄. You are definitely right, the `artifacId` does include `opencast-`, however, the artifacts uploaded to Nexus do not: (see <https://nexus.opencast.org/nexus/content/repositories/releases/org/opencastproject/>). @gregorydlogan could it be that you somehow renamed the artifacts?

gregorydlogan:
Huh, I had to modify my push script to push correctly, but I assumed that's because I had to pull it out of an old archive, not because a prefix had been dropped.  Let me diff the upstream pom files and see.

mtneug:
@gregorydlogan to be clear: the prefix is dropped in Nexus, but apparently it shouldn't. Every Opencast artifact in Nexus _should_ include the `opencast-` prefix (e.g. `opencast-common` instead of just `common`).

gregorydlogan:
Ah, I see what happened - I missed a change in my script.  I'm pushing the jars again now, give it an hour or so.

JulianKniephoff:
Alright, thanks @gregorydlogan! :smile:

@mtneug I'm going to close this then since the changes are no longer critical and, as said, you gave me an idea to make this build system at least a tiny bit better, which I'll implement soon-ish. Don't be too mad with me. :wink:

mtneug:
@JulianKniephoff not mad at all 😄 On the contrary, thanks for the great work!

@gregorydlogan it is probably a good idea to remove the wrong artifacts in order to prevent confusion.

gregorydlogan:
@mtneug oh man, yeah I suppose you're right.

PR #190, opened 2018-10-17
Translate GoldenLayout tab titles on the fly

### LINKED ISSUE #183: GoldenLayout tab titles are saved with the layout
This is a problem because of i18n. The titles are supposed to be translated to the browser language like the rest of the UI, but if the layout is saved with one language and we reload the page with another, the tab titles are still in the first language.

JulianKniephoff:
Fixes #183.


rrolf:
Although the window is not translated if I close and reopen it, I merge this PR.

Issue #159, opened 2018-08-10
Safari not working with Annotation-Tool


mliradelc:
Hi, I just tried the [annotation tool demo](https://interactivevideo.virtuos.uos.de) with safari and I have the problem that the video is not loading correctly, the video screen hangs in "start loading video" step. If I use Chrome I don't have any problems with the annotation-tool demo.

User agent details:
`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.1.2 Safari/605.1.15`

![screen shot 2018-08-09 at 16 53 11](https://user-images.githubusercontent.com/8040628/43960827-8472df80-9cb3-11e8-9538-a2458718f54c.png)

JulianKniephoff:
This is a somewhat known issue in the sense that the tool is currently mostly tested in Chrome; at least that's what I develop against. It's good to have an issue to track this stuff, the thing is just I don't know when/if I will get to it. For Safari specifically, I don't even have a (simple) way of testing at all, unfortunately. :disappointed:

On the bright side: From internal discussions I gathered that some problems just disappear when you change your user agent string to something that resembles Chrome, so you might have some luck trying that. :slightly_smiling_face:

dagraf:
LABEL: blocker
Changing the user agent string is no option for our teachers/students.

mliradelc:
Updated to Safari 12, The issue is still present.

![screen shot 2018-09-19 at 13 08 53](https://user-images.githubusercontent.com/8040628/45753514-6638c180-bc19-11e8-8991-4a0e02e04939.png)


The javascript console shows 404 that can be seen in the screenshot

JulianKniephoff:
FYI: The 404s in the console are unfortunate, but don't have anything to do with this problem.

mliradelc:
This issue was solved

PR #203, opened 2019-03-18
Use wrapper provided by MediaElement.js to listen for player events

JulianKniephoff:
Instead of listening for player control and loading events on the target element that we give to MediaElement.js we should listen to the wrapper created by it. This makes no difference for "normal" video sources, but becomes important for more exotic setups like when using HLS, where the original `video` element is not the one the user sees in the end.

### DIFF
--- a/frontend/js/player_adapter_HTML5.js
+++ b/frontend/js/player_adapter_HTML5.js
@@ -68,6 +68,7 @@ define(["jquery",
             this.initialized = false;
 
             var mediaElementPlayer;
+            var mediaElement = targetElement;
 
             /**
              * Initilize the player adapter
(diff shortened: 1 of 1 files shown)


mliradelc:
Tested and works!

mliradelc:
This PR also fixes #159, Surprisingly HLS works only with safari. Chrome and Firefox need PR #204 for HLS.

JulianKniephoff:
Interesting. I guess MediaElement is wrapping the `video` element even for "normal" videos in Safari (probably for compatibility reasons), which would explain #159. But that wrapped `video`-Element then supports HLS natively? Weird, but I take it. ¯\\\_(ツ)\_/¯

PR #216, opened 2019-04-08
Build profile in gruntfile.js works again

mliradelc:
Fixed issue where grunt file with build profile didn't work.

### DIFF
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -267,6 +267,11 @@ module.exports = function (grunt) {
             // ... the configuration
             'config': {
                 src: '<%= currentProfile.config %>',
+                dest: '<%= currentProfile.target %>/js/annotation-tool-configuration.js'
+            },
+            // ... build the configuration
+            'config-build': {
+                src: '<%= currentProfile.config %>',
                 dest: '<%= tempDir %>/js/annotation-tool-configuration.js'
             },
             // ... code for further processing
@@ -438,7 +443,7 @@ module.exports = function (grunt) {
     grunt.registerTask('default', ['jshint:all', 'less', 'copy:local-all', 'copy:local-index']);
     grunt.registerTask('baseDEV', ['handlebars:all', 'less', 'copy:all', 'processhtml:dev', 'copy:less', 'copy:config', 'copy:locales', 'concurrent:dev']);
     grunt.registerTask('baseDEMO', ['mkdir:demo', 'handlebars:all', 'less', 'copy:demo', 'processhtml:dev', 'copy:config', 'copy:locales']);
-    grunt.registerTask('baseBUILD', ['jsdoc', 'handlebars:temp', 'less', 'copy:build', 'processhtml:build', 'copy:config', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
+    grunt.registerTask('baseBUILD', ['jsdoc', 'handlebars:temp', 'less', 'copy:build', 'processhtml:build', 'copy:config','copy:config-build', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
     grunt.registerTask('baseINTEGRATION', ['handlebars:all', 'less', 'copy:integration', 'processhtml:dev', 'copy:config', 'copy:locales']);
     grunt.registerTask('baseINTEGRATIONMINIFIED', ['handlebars:temp', 'less', 'copy:integration', 'processhtml:build', 'copy:config', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
 
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -267,7 +267,7 @@ module.exports = function (grunt) {
             // ... the configuration
             'config': {
                 src: '<%= currentProfile.config %>',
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
But do all the other build tasks still work? From my understanding of the `Gruntfile`, they might not. :thinking:

`build` is not the only task that calls the `copy:config` task.

mliradelc:
Yes, that was the same assumption that I had but the profiles dev and integration work, and if the other profiles use the call `copy:config` that would have been a problem.

JulianKniephoff:
Both versions fail for me with

```
require.js:8 Uncaught Error: Script error
http://requirejs.org/docs/errors.html#scripterror
    at J (require.js:8)
    at HTMLScriptElement.onScriptError (require.js:29)
```

in the browser console, which is expected since `annotation-tool-configuration.js` is not in the right place for these tasks.

Note that you might have to get rid of previous build artifacts (the build process unfortunately does not clean these up) or just make a fresh clone of the repository.

mliradelc:
@JulianKniephoff Yes! you have right, my build artefacts where doing false good compilation, I cleared and I found the problem. Now I created a new task independent of `copy:config` called `copy:config-build`, with that I tested the dev integration and the build and the three are working.

frontend/Gruntfile.js
```diff
@@ -438,7 +443,7 @@ module.exports = function (grunt) {
     grunt.registerTask('default', ['jshint:all', 'less', 'copy:local-all', 'copy:local-index']);
     grunt.registerTask('baseDEV', ['handlebars:all', 'less', 'copy:all', 'processhtml:dev', 'copy:less', 'copy:config', 'copy:locales', 'concurrent:dev']);
     grunt.registerTask('baseDEMO', ['mkdir:demo', 'handlebars:all', 'less', 'copy:demo', 'processhtml:dev', 'copy:config', 'copy:locales']);
-    grunt.registerTask('baseBUILD', ['jsdoc', 'handlebars:temp', 'less', 'copy:build', 'processhtml:build', 'copy:config', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
+    grunt.registerTask('baseBUILD', ['jsdoc', 'handlebars:temp', 'less', 'copy:build', 'processhtml:build', 'copy:config','copy:config-build', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
     grunt.registerTask('baseINTEGRATION', ['handlebars:all', 'less', 'copy:integration', 'processhtml:dev', 'copy:config', 'copy:locales']);
     grunt.registerTask('baseINTEGRATIONMINIFIED', ['handlebars:temp', 'less', 'copy:integration', 'processhtml:build', 'copy:config', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
```
JulianKniephoff: You will probably also want to use the `copy:config-build` task here, right? Although, to be honest I don't even know whether `integrationminified` even works at the moment. :stuck_out_tongue_closed_eyes:

mliradelc: Well I haven't tried the Integration minified profile yet 😅, but with the 'uglify' process in the build the annotation tool loads way more faster.

frontend/Gruntfile.js
```diff
@@ -438,7 +443,7 @@ module.exports = function (grunt) {
     grunt.registerTask('default', ['jshint:all', 'less', 'copy:local-all', 'copy:local-index']);
     grunt.registerTask('baseDEV', ['handlebars:all', 'less', 'copy:all', 'processhtml:dev', 'copy:less', 'copy:config', 'copy:locales', 'concurrent:dev']);
     grunt.registerTask('baseDEMO', ['mkdir:demo', 'handlebars:all', 'less', 'copy:demo', 'processhtml:dev', 'copy:config', 'copy:locales']);
-    grunt.registerTask('baseBUILD', ['jsdoc', 'handlebars:temp', 'less', 'copy:build', 'processhtml:build', 'copy:config', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
+    grunt.registerTask('baseBUILD', ['jsdoc', 'handlebars:temp', 'less', 'copy:build', 'processhtml:build', 'copy:config','copy:config-build', 'copy:locales', 'copy:temp', 'requirejs', 'uglify']);
```
JulianKniephoff: Do you really need both, `copy:config` **and** `copy:config-build`, btw? :thinking:

mliradelc: The first copy the profile to the target workspace meanwhile the second you need for the `requirejs` task

JulianKniephoff: Yes but I would think that the former is not necessary since the RequireJS task should include the config file in the resulting bundle. Could you verify that it works without and then change it?

JulianKniephoff: Note also that there is a space missing here. :wink:

JulianKniephoff:
Okay I tested the things I noted in the comments and they work. I will add these changes myself and then merge this.

Sorry for rushing this now after such a long time without any answer, but there might be a release looming on the horizon. :wink:

mliradelc:
Ok, no problem. I was surprised by the answers at this moment, Nevertheless, go ahead because right now I have to work in other parts of opencast.

JulianKniephoff:
Merged in f42f731c1d5f471b578936035655458321b5c971

PR #227, opened 2019-06-14
Raise supported Opencast version

JulianKniephoff:
We should talk about the exact scheme of Opencast versions releases of this tool should support, but what seems very clear to me is that that scheme does not start at `2.3` anymore. And it probably should not start at `< 5.0`, either, so that we can finally get rid of the `artifactPrefix` option.

### DIFF
--- a/documentation/opencast-installation.md
+++ b/documentation/opencast-installation.md
@@ -30,10 +30,6 @@ In this manual we use `<annotationtool-dir>` for the base dir of the Annotation 
 This should build the frontend, include it into the Opencast modules and copies the JARs
 to your Opencast installation.
 
-Note that if you build with `opencast.version` set to something `>= 5`, you also have to specify
-`-Dopencast.artifactPrefix=opencast` because of the transition away from the `matterhorn` name
-between versions `4` and `5`.
-
 #### As a Karaf Feature
 
 As an alternative, the Annotation Tool is also packaged as a Karaf feature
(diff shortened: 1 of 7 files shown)


mtneug:
Oh thanks for getting rid of this :D

I would opt for only supporting the currently maintained Opencast versions. At the moment, that would mean the last two (7.x and 6.x).

JulianKniephoff:
Yes, that was my thinking as well. My proposal would be:

There will be one supported release of the Annotation Tool, and the `develop` version, where "supported" means "gets bug and security fixes". So kind of like Opencast itself, but with fewer versions, since we simply don't have the capacity for anything more.

Both versions of the tool, i.e. the current release and `develop` should probably support all currently supported Opencast versions. In detail: Say Annotation Tool 2.0 is released when the supported Opencast versions are 6 and 7. Then AT 2.0 should support these (and potentially `develop`?). If Opencast then moves on to only support 7 and 8, then AT 2.x should also drop support for 6, and only support 7 and 8. Same for `develop`.

As for the release cycle: I don't think a regular release makes sense for this tool at this moment. I think we can all just coordinate that as needed/when it makes sense.

mliradelc:
I agree with this change, today we use Opencast only. and this will ease the setup to many other adopters.

mtneug:
@JulianKniephoff sounds good although I would not express a support of `develop` explicitly.

PR #245, opened 2019-07-30
New config profile for use ExtApi

mliradelc:
This PR adds a new config file that allows the annotation tool to work with the external api instead of the search index.

- To work the user who logs to the Opencast system has to have the permission to access to the external API.
- In the workflows is required to attach the XACML files to the channel, if not, the annotation tool will fail in loading.

Also Added configuration files and workflows from the default installation of opencast to ease the default setup when testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ Makefile
 /frontend/doc
 
 .flattened-pom.xml
+
+#Ignore VScode files
+*.vscode
--- a/documentation/build-frontend.md
+++ b/documentation/build-frontend.md
@@ -20,9 +20,11 @@ Now you will need to initalize your NPM repository:
 You have now several options to build and deploy the webapp:
 
 * `demo` mode on a local webserver
-* `integration` to update the webapp within the Opencast backend
+* `integration` to update the webapp within the Opencast backend.
+* `integration_extapi` Changes the api endpoint to use to `api/events`.
+* `integrationminified` Integration, but minified the JS files for a faster load.
 * `build` to build the webapp and put in the the _target_ directory.
-* `dev` to start a webserver with grunt to test the current webapp
+* `dev` to start a webserver with grunt to test the current webapp.
 
 ## Building __demo__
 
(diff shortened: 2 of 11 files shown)


mliradelc:
I found something to make this more functional this PR and I need some advice:

To this be integrated is needed to have another grunt task than the default when you call maven. How would be the best way to integrate this option when you build?. A maven profile that is triggered when it is called on the CLI?

JulianKniephoff:
I actually had the same problem when I was integrating the tool with SWITCHcast. I have some SWITCH specific code for this that I will hopefully backport soon. This will require some changes to this PR, though.

mliradelc:
> This will require some changes to this PR, though.

That's right, to this to work properly you need to take into account:

-  When you built, prior you have to replace the grunt config file from the provided in the Ext_api folder and change the name of the integration task. (Too complex I think)
- You can only enter the annotation-tool with a user that is allowed to use the External API. If you try with a user without the External API permissions, the annotation tool will hang at the loading phase.


The first point to be solved is modifing the POM file to enable some option to build with a different grunt task.

The second one is more complicated, I think that the Backend should ask for this information and then send to the Frontend. 

It should be like this
ExtApi <----> Annotation-Backend<---->Annotation-Frontend

With this solution, any user (Previously Allowed) could use the Annotation Tool without having the rights for the external API

JulianKniephoff:
And both of these steps I already implemented for SWITCH. :wink:

Will have to see how well they translate to the "community edition".

JulianKniephoff:
This is pretty old and has conflicts by now. As discussed earlier, the things that these changes enable are already implemented in the SWITCH fork of the tool, and will hopefully be backported at some point. I'm sorry @mliradelc, but I'll close this now. I feel bad because the last thing I want to do is discourage participation in this project. It's exactly what it would need. But that in turn is exactly the reason why I can't look at this right now. :disappointed:

mliradelc:
Hi Julian, this PR is very old, and I changed so many things that at the end we discarded this approach so long ago. It was so long, that I've forgotten completely about this.

I've sent you a couple of emails, and I've tried through your colleagues to reach out about this, but seemed that you were not interested, but don't worry, I've made a fork of this project, feel free to take any commits from there 😄.

PR #258, opened 2019-10-09
Refactoring in preparation of the vis-timeline migration

JulianKniephoff:
I'm using the opportunity of working on porting the annotation tool over to the new vis.js timeline to improve some related issues in the codebase. This should help make the migration easier, and also improve the overall quality of the code. Maybe it even fixes some bugs in passing, although I would probably create separate pull requests for that. :wink:

I'm creating this draft PR for multiple reasons:

* The stakeholders can see that work towards the timeline migration is happening, even if it is not directly related or visible.
* They and other interested parties can already test these changes and possibly find regressions or newly introduced bugs before such a big change as the timeline migration.
* The poor soul who has to review the timeline PR at some point doesn't have to review all these little changes at the same time. They (or someone else :wink:) can do it separately.

As for the content of this PR: When you look at this, don't pay too much attention to the code changes themselves (unless you care about this stuff, of course). The purpose here is rather to find out whether I accidentally broke stuff. **The changes in this PR should not change the behavior of the application.*** Also note that I am not hesitating to rewrite the history of this branch, or force-push for any other reason.

\* As said, it might "accidentally" fix a bug, in which case I won't bother to extract it into its own PR, probably. I also won't pay too much attention to the old timeline, because this PR is in preparation for a complete replacement of that component.

---

**Note**: This now contains #265 and #267.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -3,6 +3,11 @@
          (eval . (progn
                    (c-set-offset 'arglist-cont-nonempty 8)
                    (require 'grep)
-                   (add-to-list (make-local-variable 'grep-find-ignored-directories)
-                                "node_modules")))))
+                   (make-local-variable 'grep-find-ignored-directories)
+                   (dolist (dir '("node_modules"
+                                  "frontend/js/libs"
+                                  "frontend/www"
+                                  "target"
+                                  "opencast-backend/annotation-tool/src/main/resources/ui"))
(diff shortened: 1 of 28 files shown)


JulianKniephoff:
Since #259 is nearing completion, this should now be ready for review. If any more larger refactorings should become necessary, they can just go in their own PRs, before #259 is merged.

PR #262, opened 2019-10-21
Remove annotations access control from the `Annotations` collection

JulianKniephoff:
The `Annotations` collection manages the `access` property of its children based on its parent `Track`. This is unnecessary, though, since `Annotation#access` is never used for anything.

Also: If we ever **do** need it, I would prefer a getter on `Annotation`.


rrolf:
Annotations are not loaded in this patch

JulianKniephoff:
@rrolf should be fixed. Some code was prematurely deleted here, because I my many branches confused. :sweat_smile:

Issue #264, opened 2019-10-25
Creating multiple tracks in the same session creates one more track each time


JulianKniephoff:
Steps to reproduce:

- Create a track called, say, `foo`<br>There is now one `foo` track as expected
- Create another track called `bar`<br>There are now **two** `bar` tracks! :cold_sweat: 
- Create a third (or rather fourth :sweat:) track called `baz`<br>There are now **THREE** tracks called `baz` :scream: 

**EDIT**: In fact, you need to confirm the track creation with a press of the enter key, not using the button in the modal. The latter works as expected.

PR #267, opened 2019-10-25
Stop deleted tracks from reappearing when a new track is created in the same session

### LINKED ISSUE #164: Once deleted track show up when creating a new track in the same session
If a track gets deleted and in the same session a new track is created the once deleted track shows up again and is kind auf a ghost-track: It can not be deleted or altered in any ways. When reloading the whole page the deleted track disappears.

Bug replicated in Edge and Firefox.

Video which shows the Bug (in Swiss German, but I'll guess you'll get the point anyway): https://www.screencast.com/t/A8hXCpmMz2tb

JulianKniephoff:
Fixes #164.

NB to the reviewer: When you test this, you might get confused by #264, which is already fixed in #265.

### DIFF
--- a/frontend/build/profiles/integration/annotation-tool-configuration.js
+++ b/frontend/build/profiles/integration/annotation-tool-configuration.js
@@ -136,12 +136,6 @@ define(["jquery",
             CATEGORIES_PER_TAB: 7,
 
             /**
-             * The maximal number of tracks visible in the timeline at the same time
-             * @type {Number}
-             */
-            MAX_VISIBLE_TRACKS: 0,
-
-            /**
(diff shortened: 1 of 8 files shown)


rrolf:
If I delete a track, only the first annotation within the track is removed immedeately. If there is only one annotation in the track the track is also removed. 
Otherwise the track is only removed after reload.

JulianKniephoff:
@rrolf I **think** it should be fixed now.

dagraf:
Nope, it is not yet fixed.

dagraf:
Yep, is fixed. Thx!

PR #271, opened 2019-11-08
Make the export more robust

JulianKniephoff:
When no parameters for the visibility of tracks, categories and free text annotations are specified, everything is exported, instead of silently failing with a `NullPointerException`.


JulianKniephoff:
Closing for now; this does not work with structured annotations, yet. :thinking:

JulianKniephoff:
Should work now! :smile:

PR #274, opened 2019-11-18
Make the export more robust

JulianKniephoff:
See #271

### DIFF
--- a/frontend/build/profiles/integration/annotation-tool-configuration.js
+++ b/frontend/build/profiles/integration/annotation-tool-configuration.js
@@ -155,17 +155,22 @@ define(["jquery",
              * @param {Video} video The video to export
              * @param {Track[]} tracks The tracks to include in the export
              * @param {Category[]} categories The tracks to include in the export
+             * @param {Boolean} freeText Should free-text annotations be exported?
              */
-            export: function (video, tracks, categories) {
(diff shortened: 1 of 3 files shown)


frontend/build/profiles/integration/annotation-tool-configuration.js
```diff
@@ -155,17 +155,22 @@ define(["jquery",
              * @param {Video} video The video to export
              * @param {Track[]} tracks The tracks to include in the export
              * @param {Category[]} categories The tracks to include in the export
+             * @param {Boolean} freeText Should free text annotations be exportet?
```
rrolf: exporte**d**
And maybe better "free-text"

rrolf:
I get an 404 error on:
http://my-server.tld:8080/extended-annotations/videos/1252/export.csv?track=1257&category=1258&category=1262freetext=true

JulianKniephoff:
@rrolf I can't reproduce the `404`. Neither locally nor on the SWITCH servers. :thinking:

The URL you posted confuses me: Is `my-server.tld` just an example, because you don't use `localhost`, but you also (understandably) don't want to expose your public DNS name on GitHub? Or does this really show up somewhere? I know that I have a name like that in the documentation somewhere ... :thinking:

PR #276, opened 2019-11-21
Run tests in CI

### LINKED ISSUE #198: Build currently fails with Opencast versions prior to 7
Compiling the tests fails, basically because the `RestServiceTestEnv` moved to it's own module and now lives in a different package.

As a very nasty workaround, you can of course just skip the tests using the `-Dmaven.test.skip` flag, while I figure out how to make the build compatible with both versions at least for some time.

JulianKniephoff:
Since Opencast versions prior to 7 are no longer supported (by us), this enables tests in the CI.

With that, #198 becomes irrelevant, too, so this closes #198.

This PR is only for testing the CI script on Travis. If it goes through, I might just merge it myself. xP


JulianKniephoff:
Travis in the tests with `ClassDefNotFoundError`-s. This is because we are missing transitive dependencies, because the Opencast POMs served by our nexus currently lack dependency information.

See for example https://nexus.opencast.org/nexus/content/groups/public/org/opencastproject/opencast-common/7.2/opencast-common-7.2.pom

gregorydlogan:
I'm pushing 7.4 to maven central now. Do you specifically need 7.2, or is 7.4 enough?

JulianKniephoff:
@gregorydlogan thanks! :smile: 7.4 is enough; 7.2 was just what Travis happened to grab.

Issue #84, opened 2018-01-18
Include link "Help" in menu bar


ThomasSchraml:
We wrote a documentation for our customers. It would be helpful for us to put a customizable link "Help" in the menu bar, which we can edit for referring to our documentation.

PR #278, opened 2019-11-26
Clean up the menu some

JulianKniephoff:
SWITCH wanted to organize the menu bar a bit, and I have to agree that it was kind of a mess before.

I'm not saying this is perfect, but it makes things better overall, I think.

### Screenshots

Before:

image

After:

image

image

image

image

"Process" is a bit empty, but will at least still get the Excel export **at some point* (/I promise/); "Help" could also link to some documentation, which we could provide in the future. *cough

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -28,7 +28,6 @@
             <span id="video-title" class="brand"><span data-i18n="title"></span> - <%= version %></span>
 
             <ul class="nav pull-right">
-              <li><button type="button" id="about" data-i18n="menu.about"></button></li>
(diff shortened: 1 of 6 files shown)


ebbertd:
Being able to link to documentation in the help section would be very useful. As details per organisation might be different it would be good if this link could be configured.

JulianKniephoff:
@ebbertd do we have an issue for this already? If not, would you mind creating one?

ebbertd:
@JulianKniephoff See issue #84

PR #280, opened 2019-11-26
Rudimentary separation between configuration and integration

JulianKniephoff:
This splits the configuration file into actual configuration, and "integrations" (i.e. functions and settings that configure how Annotate communicates with Opencast, or if it uses `localStorage`).

The latter is not really configuration in the sense that an adopter is not supposed to change them in their build; they are rather part of the code.

I'm not too happy with the way this code is placed and loaded, etc., but at least it is independent of the configuration now, so we can move it around in the future.


JulianKniephoff:
Merge conflict is fixed. :D

rrolf:
This version does not load if I deploy it to Opencast

rrolf:
require.js:33 GET http://rolf.virtuos.uos.de:8080/annotation-tool/js/annotation-tool-configuration.js net::ERR_ABORTED 404 (Not Found)

Arnei:
Builds, but won't load when deployed to Opencast (Version 8.3). 

Console errors:
```
Uncaught TypeError: comparators[i] is not a function
    at util.js:178
```
```
Uncaught Error: Can't create config, layout not yet initialised
    at lm.LayoutManager.toConfig (goldenlayout.js:582)
```

JulianKniephoff:
Should be fixed. :)

Arnei (approved):
Now loads and works in Opencast 8.3

PR #307, opened 2019-12-20
Prevent users from selecting and editing foreign tracks

### LINKED ISSUE #282: Double clicking a "foreign" track still allows me to edit it
I should not be able to edit other people's tracks. Ideally we should prevent that in the backend, and maybe we even do that already, I'm not sure. But at least for the local version, and for usability's sake, it would still be nice if we just couldn't open the edit dialog for tracks that don't belong to us.

### LINKED ISSUE #293: Shared tracks should be in read-only mode
When I open a track of somebody else that is public or shared with administrator I can select this track and alter it. This should not be possible. In the old version of the tool I can not select tracks of others.

What I can do:

- Double clicking on the track opens the window "Edit track" where I can edit name, description and access. Hint: This bug exists also in the old version of the tool.
- I can add own annotations to this foreign track.
- Using the options on the timeline I can move and alter the duration (start/ending) of annotations. These changes do not get saved, though.

JulianKniephoff:
This should fix #293.

### DIFF
--- a/frontend/js/views/timeline.js
+++ b/frontend/js/views/timeline.js
@@ -350,20 +350,22 @@ define([
                 }
             });
 
-            function trackButtonClicked(properties) {
-                return !!this.groupHeaders[properties.group].$el
-                    .find("button")
-                    .has(properties.event.target).length;
+            function clickedOnOneOfMyTracks(properties) {
+                var track = this.tracks.get(properties.group);
+                if (!track) return false;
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
Since this PR has conflicts, and there is still a bug (see https://github.com/opencast/annotation-tool/issues/293#issuecomment-578735706), I will close this PR for now and work on a new one.

PR #310, opened 2019-12-20
Remove frontend tests for now

JulianKniephoff:
There is literally no way to run these right now,
and there hasen't been for a while,
so they apparently don't describe reality anymore, anyway.
They **do** however clutter up the results when `grep`-ing in here...


JulianKniephoff:
Merge conflict fixed.

PR #320, opened 2020-01-24
Fix Webkit error when sorting tracks

mliradelc:
After months not seeing it now I'm back to work on the project, after testing with Webkit based browsers (Chrome, New Edge, etc) I found that if you have DASH streams showed an error.

The error occurred when sort by presenter/presentation and put DASH over the top. This error doesn't happen in Safari (Its WebKit based but is the exception) or Firefox.

I need if someone can test this fix. In theory, it supposes that it will not affect non-HLS deployments.

Best Regards.

### DIFF
--- a/frontend/build/profiles/integration/annotation-tool-configuration.js
+++ b/frontend/build/profiles/integration/annotation-tool-configuration.js
@@ -299,6 +299,10 @@ define(["jquery",
                     videos.sort(
                         util.lexicographic([
                             util.firstWith(_.compose(
+                                RegExp.prototype.test.bind(/HLS/),
+                                _.property("transport")
+                            )),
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
@mliradelc can you put the error that occurs here for reference?

Any idea why it occurs? Is it just that these browsers don't support DASH? Can't MediaElement help here?

If this is really a browser difference, then I would like to see a solution that actually looks at the browser. If this is indeed our fault, we can cheat like this for all I care, but I'm also a bit skeptical about preferring HLS not only over DASH, but over everything else as well. Can we make this a bit smarter?

Of course the real solution would be to have a proper publication channel for the tool, but apparently no one wants that. :sob: Man that would solve so many problems ...

PR #327, opened 2020-02-10
Change the way annotations are displayed in the timeline, specifically those that don't have a duration

### LINKED ISSUE #299: Newly inserted annotations can not be seen
As newly inserted annotations have no length any more they can not be seen because they get totally hidden by the red line representing the playhead.

Solution:

- The minimum width of point of time annotations is as wide as the letter "W". So all newly inserted annotations are this wide.
- To differentiate between point of time annotation and time duration annotation: Point of time annotations have an anchorage to the time bar, time durations have not.
- This anchorage has the same color as the boarder of its annotation and starts not in the middle but slightly to the left.
Bildschirmfoto 2019-12-18 um 16 44 29

JulianKniephoff:
This closes #299 for now

### DIFF
--- a/frontend/js/views/timeline.js
+++ b/frontend/js/views/timeline.js
@@ -167,6 +167,13 @@ define([
             // Options for the vis timeline
             var options = {
                 height: "100%",
+                margin: {
+                    axis: 5,
+                    item: {
+                        vertical: 5,
+                        horizontal: 0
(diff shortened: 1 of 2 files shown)


JulianKniephoff:
Closing for now (and reopening as draft) since there are still open issues in #299

PR #328, opened 2020-02-11
Prevent resetting the playhead to 0:00:00 when dragging the time marker to the far right

### LINKED ISSUE #303: When dragging the playhead over the end of the timeline it jumps back to the beginning of the video
Expected behavior: It should stay at the end of the video.

luniki:
Subtract a small delta from the video's duration. This way the video does not switch back to the beginning.

Fixes #303

### DIFF
--- a/frontend/js/views/timeline.js
+++ b/frontend/js/views/timeline.js
@@ -440,7 +440,7 @@ define([
                     Math.max(
                         0,
                         Math.min(
-                            this.playerAdapter.getDuration(),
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
This works, but the fact that I could not actually consistently reproduce this made me curious, so I did a little digging myself. Apparently this is a feature of the MediaElement library that you can just turn off. I created #350 with this in mind; I think that is the cleaner solution. Thoughts @luniki?

luniki:
#350 looks even better

PR #333, opened 2020-02-14
Add comments to the CSV export

### LINKED ISSUE #225: Export: Comments on annotations are not included
In the export CSV-file the existing comments on annotations are not included - but they should. 
Possible solutions: Either... 
(a) comments get a separate column or 
(b) get written in the column «Value». 

In addition a new column called «Refers to» is needed. Here the ID of the annotation to which the comment belongs to is indicated. See attached an export from the old annotation tool working on SWITCHcast. All rows with a number in the column «Refers To» are comments on annotations.

[annotations.xlsx](https://github.com/opencast/annotation-tool/files/3156408/annotations.xlsx)

JulianKniephoff:
This adds comments and their replies to the CSV export.

This is done using the following format. For every annotation `A`, we add `n` rows to the exported CSV (i.e. in addition to the row that just describes `A`), where `n` is the number of comments and replies under that annotation. These rows repeat all the fields of the annotation `A` itself, but add a few columns describing the comment at hand.

Specifically, a column with the comment's ID is added, and if the comment is a reply, it references the ID of the comment it is a reply to, in another additional column.

Closes #225 for now.

Note that due to #330, the additional comment lines might be out of the expected order.


JulianKniephoff:
Since there is still an issue here (see https://github.com/opencast/annotation-tool/issues/225#issuecomment-590254945), I will close this for now.

JulianKniephoff:
Okay I seem to have resolved the remaining issue.

Issue #295, opened 2019-12-18
Changing the order of tracks does not work


dagraf:
Changing the order in the window "edit tracks" hast no effect.

Steps to reproduce:
- open the edit tracks window
- alter order of tracks using the arrows
- hit confirm

luniki:
I tried to reproduce this. In my browser the new position of the tracks after hitting confirm is at first correct. But as soon as I reload the browser window the positions snap back to their previous position.

luniki:
I need feedback to work on this issue. 

"Track management" works on a forest of (shallow) trees of tracks. When I select and reorder visible tracks in the "Track management" modal, what is expected to happen exactly? 

In my tests only the selected, visible tracks are shown in the order I defined. This is certainly expected. But the selection of visible tracks and the position of the visible tracks in the track list  is not persisted. Is this the problem of this issue?

AFAIK there is currently no existing method to persist this visibility filter and positioning.

dagraf:
[Reihenfolge der Spuren ändern.zip](https://github.com/opencast/annotation-tool/files/4184768/Reihenfolge.der.Spuren.andern.zip)

See screencast for how the bug looks like. First part in Chrome, second in Firefox.

luniki:
Could you add your browser versions to this issue? I cannot reproduce this in the latest FF/Chrome.

dagraf:
Chrome: Version 80.0.3987.87 (Offizieller Build) (64-Bit) 
Firefox: 72.0.2 (64-Bit)

Beides auf Windows.

luniki:
I tried this with Chrome Version 80.0.3987.100 (Official Build) (64-bit) on a Linux laptop, but there the re-ordering works like expected.
[Presentation video - Annotation Tool - Track order.webm.zip](https://github.com/opencast/annotation-tool/files/4190663/Presentation.video.-.Annotation.Tool.-.Track.order.webm.zip)

luniki:
Perhaps @JulianKniephoff has access to a Windows computer?

BTW I am using the current "master" via "npm run start"

dagraf:
Bug also appears on Mac:
- Chrome: Version 80.0.3987.100 (Offizieller Build) (64-Bit)
- Safari: Version 13.0.5 (14608.5.12)

Observation: The order is always a chronological one: Older tracks are desplayed below.
Tested on SWITCHcast and SWITCHcast-Test

luniki:
[...]

PR #335, opened 2020-02-18
fix: convert string to number for correct sorting

luniki:
The list of track IDs is sometimes a list of strings instead of integers. AnnotationTool#orderTracks sorts the tracks incorrectly when the new order is a list of strings.

Refs #295

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -434,11 +434,14 @@ define(["jquery",
              * @param {Array} order The new track order
              */
             orderTracks: function (order) {
+                // convert the new order to string to compare reliably
+                var strOrder = order.map(function (item) { return "" + item; });
                 //   Make sure every visible track is represented in the order,
                 // and only those, with non-explicitly ordered tracks in front.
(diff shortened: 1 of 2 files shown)


JulianKniephoff:
Wait, but sometimes the IDs are proper, non-numeric strings, right? The `localStorage`-"backend" uses UUIDs for example, IIRC.

luniki:
Ok, so track IDs are really strings? I will rewrite the orderTracks method to use string comparison.

luniki:
I removed the track ID to number conversion and instead converted the array containing the ordered IDs and the track IDs to string to compare them reliably.

JulianKniephoff:
Okay, thinking about this a bit more, I actually think it's a bug that the frontend ever treats entity IDs as a `Number`, and in the long run, we should fix that. I tracked #351 for this.

Of course that's a bigger issue and won't happen in the rest of the time we have for https://github.com/opencast/annotation-tool/projects/2, so I will just go ahead and merge this as a quick fix for #295.

PR #336, opened 2020-02-18
fix: change implementation of Annotation#covers according to #317

### LINKED ISSUE #317: Annotations shorter than 5 sec are to long indicated as active
Point-in-time annotations and duration annotations shorter than 5 seconds are displayed as active for 5 seconds (see https://tube.switch.ch/videos/becff00e).

Desired behavior: 
- Point-in-time annotations: Indicated as active for 1 second
- Duration annotations: Active only as long as the actual duration of the annotation

luniki:
Touching Annotation#covers changes:

* Point-in-time annotations are now indicated as active for 1 second
* Duration annotations are now indicated as active only as long as the actual duration of the annotation

Fixes #317

### DIFF
--- a/frontend/js/models/annotation.js
+++ b/frontend/js/models/annotation.js
@@ -243,8 +243,8 @@ define(["underscore",
             covers: function (time, minDuration) {
                 var start = this.get("start");
                 var duration = this.get("duration");
-                var end = duration === 0 ? start + minDuration : start + end;
-                
+                var end = start + (duration || minDuration);
+ 
                 return start <= time && time <= end;
             },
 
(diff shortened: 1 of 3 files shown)


frontend/js/models/annotation.js
```diff
@@ -242,8 +242,9 @@ define(["underscore",
              */
             covers: function (time, minDuration) {
                 var start = this.get("start");
-                var duration = Math.max(minDuration, this.get("duration") || 0);
-                var end = start + duration;
+                var duration = this.get("duration");
+                var end = duration === 0 ? start + minDuration : start + end;
```
JulianKniephoff:
```suggestion
                var end = start + (duration || minDuration);
```

? :thinking:

luniki: Sure.

PR #338, opened 2020-02-19
Add a border below the last track

### LINKED ISSUE #292: Insert a line between the lowest track
At the moment, there are lines between tracks but no line to mark the lower end of the lowest track. Please insert a line to mark this lower end of the lowest track.

JulianKniephoff:
The timeline displays a thin line to separate tracks. It would look better and more consistent if this line also "bookended" the last track. This PR adds that.

Fixes #292.

### DIFF
--- a/frontend/style/annotations/timeline.less
+++ b/frontend/style/annotations/timeline.less
@@ -2,6 +2,13 @@
     background-color: white;
     display: flex;
     flex-basis: 0;
+
+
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
I thought I remember an issue that asked for this even, but I can't find it now. If anyone sees it, just let me now here. :smile:

JulianKniephoff:
Found it!

Fixes #292.

PR #339, opened 2020-02-20
Properly resize the GoldenLayout when the window is resized

### LINKED ISSUE #291: Timeline controls get hidden when the screen gets downsized
When the screen gets downsized the timeline controls disappear.

JulianKniephoff:
Fixes #291.

### DIFF
--- a/frontend/js/views/main.js
+++ b/frontend/js/views/main.js
@@ -389,9 +389,11 @@ define(["jquery",
                     // in order to prevent the play-pause shortcut (space)
                     // to be handled twice.
                     self.playerContainer = container.getElement()[0];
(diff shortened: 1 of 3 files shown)


frontend/js/views/main.js
```diff
@@ -805,6 +807,8 @@ define(["jquery",
              * @alias module:views-main.MainView#onWindowResize
              */
             onWindowResize: function () {
```
luniki: It would be nice to debounce "onWindowResize" using underscore's [debounce](https://github.com/jashkenas/underscore/blob/1.8.3/underscore.js#L813) function to prevent sluggish behaviour.

PR #340, opened 2020-02-20
Fix converting range annotations to point-in-time

### LINKED ISSUE #300: Shortening a time duration annotation to a point of time annotation does not work
Trying to shorten a time duration annotation to a point of time annotation by dragging the end/beginning of the annotation in the timeline does not work. On the timeline the annotation appears as a point in time annotation but in the list of annotation the annotation has still the old start- and end-time indicated.

When pulling the start of an annotation to the end this start can be pulled further than the actual end of the annotation which ends in an point in time annotation with a time that is later in time that the initial end of this annotation. In the list of annotation this annotation is still indicated as time duration annotation with the initial start and end time.

JulianKniephoff:
Fixes #300.

Note that this includes #332.


frontend/js/views/timeline.js
```diff
@@ -76,17 +76,12 @@ define([
...
-                "color:" + (
-                    chroma(color).luminance() < 0.5
-                        ? "white"
-                        : "black"
-                )
-            ].join(";");
+            item.className = "category-" + label.category.id;
         }
-        item.type = "range";
+        console.log(item.duration);
```
luniki: Debugging console.log left in the code...

frontend/js/views/timeline.js
```diff
@@ -545,6 +546,34 @@ define([
...
+                            ".vis-item.vis-selected.category-" + category.id + "{" +
+                            "background-color:" + color + ";" +
+                            "color:" + (
+                                chroma(color).luminance() < 0.5
+                                    ? "white"
+                                    : "black"
+                            ) +
+                            ";}";
+                            }).join("");
+                return $("<style>" + stylesheet + "</style>")
```
luniki: Why don't you use .insertRule instead of this?

JulianKniephoff: Because I didn't think of it, tbh. :wink: 

But also https://caniuse.com/#search=insertrule worries me a bit. Thoughts?

JulianKniephoff:

PR #342, opened 2020-02-25
fix: use different icons for track visibility

### LINKED ISSUE #341: Omit eye icon for "access" settings, becaus we us the eye icon differently now (hide annotations). Idea:
Public = Globe
Private = Lock or person-icon

luniki:
Fixes #341

### DIFF
--- a/frontend/style/annotations/base_style.less
+++ b/frontend/style/annotations/base_style.less
@@ -53,8 +53,8 @@ body {
 .icon-likert-plusplus    { background-position: -336px -168px !important; }
 .icon-color-palette      { background-position: -360px -168px !important; }
 .icon-color-corner       { background-position: -384px -168px !important; }
-.icon-public             { background-position: -336px -144px !important; height: 15px !important; width: 23px !important; }
-.icon-private            { background-position: -288px  -24px !important; height: 15px !important; width: 23px !important; }
+.icon-public             { background-position: -336px -144px !important; }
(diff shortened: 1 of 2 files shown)


JulianKniephoff:
Whupps, accidentally deleted the wrong branch. :sweat_smile:

frontend/style/annotations/base_style.less
```diff
@@ -53,8 +53,8 @@ body {
 .icon-likert-plusplus    { background-position: -336px -168px !important; }
 .icon-color-palette      { background-position: -360px -168px !important; }
 .icon-color-corner       { background-position: -384px -168px !important; }
-.icon-public             { background-position: -96px  -120px !important; height: 15px !important; width: 23px !important; }
-.icon-private            { background-position: -120px -120px !important; height: 15px !important; width: 23px !important; }
```
JulianKniephoff: With this change, we should probably get rid of these custom `width` and `height` settings. These were only there since the old private icon was a bit wider than the rest.

Currently this makes the icons in the "access popover" weirdly spaced.

Without the `height` in particular, that popover will probably also need a `vertical-align: center;`. :thinking:

luniki: I removed the width and height. The "vertical-align: center" wasn't necessary but the removal of the default "margin-top: 1px" (actually) centering the icons vertically in the pop-up menu.

PR #346, opened 2020-02-25
fix: update vis-timeline library to v7.1.3

### LINKED ISSUE #286: Cursor: Change cursor to "hand" when annotations/playhead can be moved
At the moment the cursor changes from an arrow to a cross with arrow endings when hovering over the playhead or over a selected annotation.
1. Change should be to a "hand" and not a cross.
2. Cursor should also change when hovering over a not active annotation (as it was the case with the old timeline).

luniki:
Fixes #286


JulianKniephoff:
This needs 7661a050dcd31f7da48e1a9d4c91240a7a595d60 from #332 to work correctly. Otherwise the alignment of the labels within ranged annotations is wrong.

Other than that I can't find any issues caused by the update right now. I also tested this together with #332, and it seems to work fine, but a second pair of eyes is of course appreciated. I think we should merge #332 first, and then this can go in.

PR #348, opened 2020-02-26
fix: debounce window resize handlers

luniki:
window resize handlers should always be debounced to prevent sluggish behaviour.

### DIFF
--- a/frontend/js/views/annotate-category.js
+++ b/frontend/js/views/annotate-category.js
@@ -146,7 +146,8 @@ define(["jquery",
                     this.listenTo(annotationTool, annotationTool.EVENTS.ANNOTATE_TOGGLE_EDIT, this.switchEditModus);
                 }
 
-                $(window).on("resize.annotate-category", this.updateInputWidth);
+                var onResize = _.debounce(this.updateInputWidth, 250);
+                $(window).on("resize.annotate-category", onResize);
 
                 //this.render();
                 this.nameInput = this.$el.find(".catItem-header input");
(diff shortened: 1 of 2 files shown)


luniki:
The "magic number" is situationally and I do not think that there is a single number that fits all.

I decreased the debounce time to 30ms in 4c63aab which results IMHO in a better jitter free resize handling with a better FPS.

If you'ld like, I can make some flame charts to document the performance gains.

JulianKniephoff:
> The "magic number" is situationaly and I do not think that there is a single number that fits all.

You are right, that's actually a good point. :thinking:

> I decreased the debounce time to 30ms in [4c63aab](https://github.com/opencast/annotation-tool/commit/4c63aab302f6d191baf748036afbcf40951d2046) which results IMHO in a better jitter free resize handling with a better FPS.
> 
> If you'd like, I can make some flame charts to document the performance gains.

I don't doubt that this increases the performance; it almost tautologically does less work in the same amount of time after all. I'm only arguing that the noticable delay we get from debouncing might just feel like another kind of lag to the user (it does to me). And that is a lag that we unconditionally introduce to anyone, not just those that have machines that are too slow to run the resize handlers and the corresponding reflows 60 times a second.

Would be curious to see whether there is some kind of adaptive solution to this. :thinking:

Anyway, I'll see how I like it with the smaller delay and report back. x)

JulianKniephoff:
Okay I had another look, but I have to say, I still don't like it too much. I feel bad, but I think I will have to decline this. :thinking: I'm sorry! :disappointed_relieved: 

If we really want this in the long run, I would want to have it adapt to the user, either automatically by somehow measuring how much lag the user experiences, or manually using a config option somehwere. :thinking:

PR #349, opened 2020-02-26
Don't round annotation start and end times

### LINKED ISSUE #313: Setting of end and start times of annotation are not accurate
When altering the length of an annotation with a duration using the start and end buttons the new start or end of an annotation is not perfectly accurate. It is slightly earlier or later than the actual position of the playhead (see screencast: https://tube.switch.ch/channels/9a280238).

### LINKED ISSUE #318: Annotations are not placed on the exact position of the playhead
Annotations are inserted on the nearest second and not on the exact playback time. This means that the desired playback time is not displayed when selecting the annotation. See tube.switch.ch/videos/fef0e56b

JulianKniephoff:
* Fixes #318 
* Fixes #313

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -510,7 +510,7 @@ define(["jquery",
                 var annotation = this.selectedTrack.annotations
                     .create(_.extend(
                         params,
-                        { start: Math.round(this.playerAdapter.getCurrentTime()) },
+                        { start: this.playerAdapter.getCurrentTime() },
                         // The loop controller can constrain annotations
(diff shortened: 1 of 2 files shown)


dagraf:
And what about https://github.com/opencast/annotation-tool/issues/313? Does this pull request also fix this issue?

JulianKniephoff:
@dagraf I don't think so unfortunately. :frowning_face:

JulianKniephoff:
Now it does. :wink:

dagraf:
FANTASTIC!! :)

JulianKniephoff:

PR #354, opened 2020-02-28
Change the way selections work

### LINKED ISSUE #331: Single click selects annotation, double-click jumps to it
When I have freshly selected a time-in-point annotation (that has now an anchor) and I want to move this annotation I have to almost click on the edge of the annotation to be able to do so. Otherwise I move the playhead. This is because the red line of the playhead "takes to much space" in the sense of when clicking slightly right or left of the playhead I still move the playhead and not the annotation below.

Solution: Make this space much smaller (at least half as wide as now).

### LINKED ISSUE #356: Free text annotations: Written text remains in box
Since the last Update of SWITCHcast-Staging: After writing a free text and hitting «Enter», the annotation gets inserted (as expected) and the written text remains in the text box (not expected). The text box should be empty after having inserted an annotation.

JulianKniephoff:
This decomplects the selection (one annotation, clicked on by the user in the list or the timeline) from "active annotations" (the ones that are covered by the play head, currently).

While doing so, a lot of selection related code was refactored as well.

---

It also closes #331, which was easy to do now.

2c75976 now also closes #356.


frontend/js/views/list.js
```diff
@@ -243,90 +239,104 @@ define(["underscore",
...
-                            selectedAnnotations[i] = view;
-                        }
-                    }
+            renderSelection: function (selection, previousSelection) {
+                if (previousSelection) {
+                    this.getViewFromAnnotation(previousSelection.id)
+                        .$el.removeClass("selected");
                 }
+                if (selection) {
+                    var view = this.getViewFromAnnotation(selection.id).$el;
```
Arnei, line 253: Creating a freetext annotation produces a type error: 
`TypeError: this.getViewFromAnnotation(...) is undefined`
Doesn't seem to hamper functionality, but may be worth fixing nonetheless.

Arnei:
A "selected" annotation is not visually marked as an "active" annotation, even if it is covered by the play head. Is this intended? (I assume it is, but it had me confused for a while, so I thought I'd ask.)
Peek 2020-06-18 15-29

JulianKniephoff:
The commit I just pushed should fix #356.

frontend/js/views/list.js
JulianKniephoff: Actually this was exactly the cause for #356. Should be fixed by 2c75976 now.

JulianKniephoff:
> A "selected" annotation is not visually marked as an "active" annotation, even if it is covered by the play head. Is this intended? (I assume it is, but it had me confused for a while, so I thought I'd ask.)

It was intended by me, because selection, in my mind, is somehow "stronger" than just being active. However, you kind of hit a nerve here; there was lots of discussion about this in #359 and I intend to resolve this during the finishing touches of the Timeline project, but I think that merits another PR.

PR #386, opened 2020-05-15
Set minimal zoom of the time line to at least the video duration

### LINKED ISSUE #385: The timeline shows "invalid" regions when the video is very short
As can be seen here:

image

As you can see in the player, this is a one second video, but the timeline shows five, two of which are negative.

JulianKniephoff:
Fixes #385.


JulianKniephoff:

PR #404, opened 2020-07-08
Visualize selectable annotations better

### LINKED ISSUE #286: Cursor: Change cursor to "hand" when annotations/playhead can be moved

JulianKniephoff:
Closes #286

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -395,7 +395,7 @@ define(["jquery",
                     ), {
                         wait: true,
                         success: _.bind(function () {
-                            this.setSelection(annotation, true);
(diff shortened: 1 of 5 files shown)


JulianKniephoff:
@Arnei I agree and I changed it accordingly. I also did some more tweaking. The annotations in the list now also get dimmed when hovering over them, like the timeline items, which makes the whole thing even more consistent.

Note to self: This now includes #406, so that should probably be merged first.

PR #406, opened 2020-07-08
Make the distinction between active and selected annotations more consistent and less confusing

### LINKED ISSUE #359: Different colours for selected and activated annotations is confusing (and not consistently implemented)
It is hard to understand which coulour means what. In the end it confuses more than it helps. I suggest we only use the old version of highlighting active annotations and ged rid of the new one.

Here's whats not consistent:
- Single click= red
- Double click = red & Playhead jump to this position (correct would be YELLOW)
- Single click > double click on the same annotation = yellow & Playhead jump to this position

JulianKniephoff:
This should address all the remaining issues of, and thus close #359.

Specifically:

* Double clicking annotations in either the timeline or the list view now unconditionally selects the annotation, no matter it's previous state
* This is not a change, but to clarify: Double clicking also brings the play head to the corresponding annotation, ...
* while single clicking still doesn't; it only selects/deselects an annotation (depending on the previous state)
* Active annotations got their red "border" back \o/
* The selected annotation is now cross-hatched in the background

I hope I remembered our compromise correctly, @dagraf; feel free to correct me.

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -104,8 +104,8 @@ define(["jquery",
 
             /**
              * Initialize the tool
-             * @alias   annotationTool.start
-             * @param  {module:annotation-tool-configuration.Configuration} config The tool configuration
+             * @alias annotationTool.start
+             * @param {module:annotation-tool-configuration.Configuration} config The tool configuration
              */
             start: function (config) {
                 _.bindAll(this,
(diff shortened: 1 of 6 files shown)


JulianKniephoff:
@Arnei you are right, I can reproduce that relatively consistently, and it seems to be a browser issue: The browser, or I should say my browser, only seems to group clicks up to 3. So a quadruple-click is interpreted as two double clicks, a single click and a triple click, or a triple click and a single click. The first two possibilities are fine, but the latter would deselect the annotation.

The timeline does not seem to have this problem, interestingly enough. They probably implemented some browser independent logic, which I would rather like to avoid. Maybe we can look for a lib to do that for us at some point. :thinking: 

So yes, it bugs me, but I don't think this is a big deal in the context of this PR. @Arnei would you mind creating an issue?

PR #408, opened 2020-07-27
Added an option to export statistics to Excel (XLXS)

Arnei:
This PR adds an additional button to the "Process" menu item, which allows to download statistics in an excel-compliant format to work with Excel applications that cannot parse CSV, such as Excel Online. Related to issue #20 .

- Excel-Export happens in the frontend, using a new js library.
- Also pulled CSV-Export into the frontend, using a new js library.
- Also removed CSV-Export from the backend


frontend/js/util.js
```diff
@@ -248,6 +248,18 @@ var util = {
...
+
+    /**
+     * Capitalize a string
+     * @param {*} s The string to capitalize
+     * @return {s} The capitalized string
+     */
+    capitalize: function (s) {
+        if (typeof s !== 'string') {
+            return '';
+        }
```
JulianKniephoff, line 261: If we bother to type check here at all, I would like this to fail louder (i.e. `throw` an `Error`) or be nicer by just returning `s` when it's not a string, so as to not just "swallow" input randomly.

But honestly I wouldn't bother. If people call this with something other than a string, the next line will fail anyway, and they shouldn't do that anyway. :wink: 

```suggestion
```

frontend/js/util.js
```diff
@@ -248,6 +248,18 @@ var util = {
         return _.extend(object, _.object(_.map(properties, function (property) {
             return [property, definition(property)];
         })));
+    },
+
+    /**
+     * Capitalize a string
+     * @param {*} s The string to capitalize
+     * @return {s} The capitalized string
```
JulianKniephoff, line 256: This should be

```suggestion
     * @return {string} The capitalized string
```

frontend/js/util.js
```diff
@@ -248,6 +248,18 @@ var util = {
         return _.extend(object, _.object(_.map(properties, function (property) {
             return [property, definition(property)];
         })));
+    },
+
+    /**
+     * Capitalize a string
+     * @param {*} s The string to capitalize
```
JulianKniephoff, line 255: With my suggestion below, this should be

```suggestion
     * @param {string} s The string to capitalize
```

frontend/locales/de/translation.json
```diff
@@ -126,7 +126,8 @@
     },
     "processing": {
       "head": "Verarbeiten",
-      "export": "Für Statistik exportieren",
+      "export_csv": "Für Statistik exportieren",
+      "export_xlxs": "Für Statistik exportieren (Excel)",
```
JulianKniephoff, line 130:
```suggestion
      "export csv": "Für Statistik exportieren",
      "export xlxs": "Für Statistik exportieren (Excel)",
```

The convention followed for translation keys so far was to not put any delimiters between words.

frontend/build/profiles/integration/annotation-tool-configuration.js
```diff
@@ -151,20 +153,194 @@ define(["jquery",
...
+                    bookData.push(commentLine);
+                }
+
+                function comment_replies(line, replies) {
+                    _.each(replies, function (comment) {
+                        addCommentLine(line, comment);
+
+                        comment_replies(line, comment.attributes.replies)
+                    });
+                }
```
JulianKniephoff, line 343: Now that the export functionality is no longer dependent on whether or not we are running with the backend or in "`localStorage`-mode", I think these functions don't belong here, anymore. I would rather see them in `annotation-tool.js`.

frontend/build/profiles/integration/annotation-tool-configuration.js
```diff
@@ -151,20 +153,194 @@ define(["jquery",
              * @param {Category[]} categories The tracks to include in the export
              * @param {Boolean} freeText Should free-text annotations be exported?
              */
-            export: function (video, tracks, categories, freeText) {
-                var parameters = new URLSearchParams();
+            export_csv: function (video, tracks, categories, freeText) {
```
JulianKniephoff, line 156: Function and variable names in JavaScript should be in `camelCase`.

```suggestion
            exportCSV: function (video, tracks, categories, freeText) {
```

(Note: Whether it should be `exportCsv` or `exportCSV` is debatable. I prefer the latter, because CSV is an acronym, so in a way, each letter is its own word.)

(Also note that I won't mention these kinds of style and naming issues on every instance; you are invited to look for potential further places to fix them yourself. x))

JulianKniephoff (changes requested):
@Arnei I'm not done yet, but here are some things I already found while trying to bring this to the SWITCH staging system, just in case you already find some time to look at them before I continue this review tomorrow. :wink:

documentation/Rest-video.md
```diff
@@ -89,20 +89,3 @@ NO CONTENT
...
-
-### Example request
-#### _Url_
-**GET** `http://api.annotationstool.com/v1/videos/#{id}/export.csv `
-#### _Content_
-`access=public`
-#### _Response content_
-`text/csv`
-
-[< Rest API documentation](Rest-API.md)
```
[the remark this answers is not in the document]
Arnei: Nope, good catch.

frontend/build/profiles/integration/annotation-tool-configuration.js
Arnei: (And I've switched to the former, as that appears more readable to me when the acronym is followed by another word. )

(These reminders are still very appreciated :bow: )

frontend/js/annotation-tool.js
```diff
@@ -685,9 +689,212 @@ define(["jquery",
...
+                            line.push(label.value)
+                            line.push(label.abbreviation)
+                        } else {
+                            line.push("")
+                            line.push("")
+                            line.push("")
+                        }
+                        
+                        // What the heck is the difference between 'scalevalue' and 'scaleValue'?
+                        // They seem to contain the exact same data
```
JulianKniephoff, line 816: Let me introduce you to one of the biggest monstrosities of this codebase: https://github.com/opencast/annotation-tool/blob/master/frontend/js/models/annotation.js#L89-L92

And then later on in the same function: https://github.com/opencast/annotation-tool/blob/master/frontend/js/models/annotation.js#L129-L131

I have no idea why. :sob: 

```suggestion
```

frontend/js/annotation-tool.js
```diff
@@ -685,9 +689,212 @@ define(["jquery",
...
+                        if(annotation.attributes.scalevalue) {
+                            line.push(annotation.attributes.scalevalue.scale.name);
```
JulianKniephoff, line 818: Unfortunately, scale values apparently don't always have this convenient back link to their parent scale. This errors in `localStorage`-mode.

JulianKniephoff (changes requested):
Okay, here is the rest of my review. I only tested this in `localStorage`-mode for now, where it didn't work for me, because of some of the comments I added.

frontend/js/annotation-tool.js
```diff
@@ -685,9 +689,212 @@ define(["jquery",
...
+                header.push("Comment text");
+                header.push("Comment replies to");
+                bookData.push(header);
+
+                _.each(tracks, function (track) {
+                    _.each(annotationTool.getAnnotations(track.id), function (annotation) {
+                        var line = [];
+
+                        let label = annotation.attributes.label;
+                        // No idea what this check is good for tbh
```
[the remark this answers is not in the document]
Arnei: Well now this is embarrassing. Thanks for even answering questions I ask myself in the comments.

frontend/js/annotation-tool.js
Arnei: ...Fascinating, thank you.

JulianKniephoff:
Note to self: There might be some more dependencies we can remove from the backend bundles, now that the CSV export is in the frontend.

PR #410, opened 2020-08-21
Preserve annotation anchor coloring even for non-selected, active annotations

### LINKED ISSUE #358: Anchor of label annotations: When activated via single click they turn black
At least the dot on the timeline should keep its colour.

JulianKniephoff:
This is an amendmend to #398, and should address @dagraf's [comment](https://github.com/opencast/annotation-tool/issues/358#issuecomment-658118100) on #358. It should thus close #358.

### DIFF
--- a/frontend/style/annotations/base_style.less
+++ b/frontend/style/annotations/base_style.less
@@ -192,10 +192,11 @@ input, textarea, select, .uneditable-input {
     }
 }
 
-@annotationSelectionColor: rgba(0, 0, 0, 0.125);
+@annotationSelectionColor: rgba(0, 0, 0, 0.15);
 
 .selectable-annotation {
     cursor: pointer;
+    color: black;
     &:hover {
         background-image: linear-gradient(@annotationSelectionColor, @annotationSelectionColor);
     }
--- a/frontend/style/annotations/list.less
+++ b/frontend/style/annotations/list.less
@@ -174,7 +174,6 @@
         .start,.end {
             display: inline-block;
             position: relative;
-            color: @black;
 
             &.no-duration {
                 display: none;
@@ -187,7 +186,6 @@
                 margin: 0;
                 padding: 0;
                 box-shadow: none;
-                color: @black;
                 text-align: center;
                 font-family: monospace;
                 background: transparent;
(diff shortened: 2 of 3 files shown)


frontend/style/annotations/base_style.less
```diff
@@ -213,4 +214,5 @@ input, textarea, select, .uneditable-input {
         @annotationSelectionColor 10px,
         @annotationSelectionColor 20px
     );
+    .selectable-annotation();
```
Arnei, line 217: Maybe it would be better to also use a repeating-linear-gradient for the hover of the selected-annotation, instead of the default one? I think the selected annotation should always stay striped.

JulianKniephoff: I think I agree with that. :thinking: :sweat_smile: 

Will see what I can do.

JulianKniephoff: Right, now I remember why I did this: This way you have no indication that clicking an already selected annotation would do anything. I guess having the cursor still be the pointy hand is kind of a hint, but on the timeline you don't even have that. :thinking: 

I changed it back for now, because being clear about what is selected is probably more important than the above, but I might revise this in the future, if I can think of a better solution.

Arnei: So if I'm understanding you correctly, you would like the hover to be a preview of what happens when you click the annotation? E.g. when you hover a striped (active) annotation, the hover should be non-striped (not-active)? That seems unnecessary to me.

Also I just tested this, and the hover for a selected annotation is still non-striped.

JulianKniephoff: I don't exactly know what I want, yet. :sweat_smile: I would just like to have some indication that clicking a selected annotation does something.

Have you pulled the latest commits? I force-pushed, so maybe you have to do `git`-magic. :see_no_evil:

Arnei: Nvm, I forgot to fetch, my bad. 
The selected annotation now retains its stripes, but does not get highlighted when hovered, which also seems sub-optimal. But since this problematic is not directly related to the topic of the PR, I'll consider this resolved.

Arnei (approved):

PR #413, opened 2020-09-14
Drop some questionable compatibility layers

JulianKniephoff:
If we ever want to have any hope to bring this tool into the modern age of web development, it's time to drop some unmaintained, untested and otherwise questionable libraries to maintain compatibility with unsupported versions of Internet Explorer.

We could probably do more, but these stood out to me while looking at #408. This also removes some completely unused libraries and does some cleanup around the places it touches.

For the reviewer: Functionality-wise this should only affect import and export of categories, so that's probably what should be tested most extensively.


Arnei (approved):
Works in Iridium 2020.04 and Firefox 74.0.

Pretty straightforward PR.

PR #431, opened 2020-12-14
Get rid of the meaningless version

### LINKED ISSUE #312: Get rid of the meaningless version
The tool still contains references to a version a la 3.0.0. This is now meaningless with hour new "release" process.

I saw it at least in:

* `package.json`
* which is in turn used by Grunt somehow to put it in `version.js`
* Which is used in `about.js`

JulianKniephoff:
This also removes the "About" dialog, which basically only displayed
this, and the the "Help" menu, which only held that dialog, in turn.

Closes #312.


frontend/index.html
```diff
@@ -92,9 +92,6 @@
                 <button type="button" class="dropdown-toggle" data-toggle="dropdown">
```
Arnei, line 92: The menu button still remains:
Bildschirmfoto vom 2020-12-16 12-15-31
(Deployed in Opencast)

The button does not do anything. It may be enough to just remove the whole dropdown in the index.html

Arnei (changes requested):
The menu button still remains:
Bildschirmfoto vom 2020-12-16 12-15-31
(Deployed in Opencast)

The button does not do anything.

frontend/index.html
JulianKniephoff: Should be fixed. :)

Arnei (approved):
Can confirm, no more button

Issue #425, opened 2020-12-10
Bern features: Changing the color of a category has no immediate effect on annotations shown in the list


dagraf:
When I change the color of a category all annotations made with this category...

- in the timeline: ... immediately change their color.
- in the list of annotations: ... only change their color after a browser reload.

Solution: Also the annotations in the list should immediately change their color.

dagraf:
It seems to be different of what I described above:
- in the timeline: ... their color change immediately.
- after a reload: ... the category and all its annotations (in the timeline and in the list) are shown again with the old color.

It seems to me that the change of color is not properly saved and is lost after a reload.

dagraf:
Same problem (changes are not properly saved) appears when changing a title of a category.

dagraf:
Looks fine and can be closed. Thx!

PR #434, opened 2021-01-11
Set color and background color as inline style in the timeline

### LINKED ISSUE #427: Bern features: Annotation of categories belongig to a series have no color in "daughter tracks"
If I create a category belonging to a series in video A and then in video B click on a label of this category, the inserted annotation has NO color on the track (meaning = is grey). In the list of annotations the annotation has the correct color. Using the same category in video A (where it got created) does not lead to such missing coloring on the track.

Solution: Please make sure that annotations always are correctly colored regardless of which video the category was created in. 

Screenshot: Categories E3 and S3 were created in this video and their annotations are correctly colored on the track. Categories S1 and S2 were created in other videos of the same series and their annotations are grey instead of orange or yellow.
Bildschirmfoto 2020-12-10 um 13 11 38

JulianKniephoff:
This is probably slightly less efficient, but makes some upcoming
changes simpler. Also the stylesheet solution was a bit janky to begin
with, and we should probably touch that code again in the long run
anyway.

This should fix #427 when integrated in #414.

### DIFF
--- a/frontend/js/models/annotation.js
+++ b/frontend/js/models/annotation.js
@@ -250,12 +250,31 @@ define(["underscore",
 
             /**
              * Access an annotations category, if it has any.
+             * Note that this returns <code>undefined</code>
+             * if the category has been deleted!
              * @alias module:models-annotation.Annotation#category
              * @return {Category} The category this annotations label belongs to, if it has a label
              */
             category: function () {
                 var label = this.get("label");
                 return label && annotationTool.video.get("categories").get(label.category.id);
+            },
+
+            /**
+             * Get the display color of an annotation.
+             * This is determined by the color of the category of its label,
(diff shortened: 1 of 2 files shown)


Arnei (commented):
Seems to solve the intended issue, great.

Also makes annotations in the timeline not immediately change their color anymore, when the color of their category changes. Likely related to #425.

JulianKniephoff:
Right, that was actually one of the advantages of the stylesheet approach. :thinking: 

I guess I could just adapt what I did for #435. Let me see ...

JulianKniephoff:
Note that this now uses #437, which should probably reviewed first.

Arnei (approved):
Approving this by virtue of being unable to break it.

PR #435, opened 2021-01-12
Update the color of structured annotations in the list view ...

JulianKniephoff:
... as soon as the color of a category changes.

Note: This is a quick fix that we need for some upcoming changes,
and the quality of the code surrounding these changes does not warrant
a clean solution right now.

Let's just hope we will get to it one day.

This should solve at least part of #425.

### DIFF
--- a/frontend/js/views/list-annotation.js
+++ b/frontend/js/views/list-annotation.js
@@ -366,7 +366,10 @@ define(["jquery",
                     category = annotationTool.video.get("categories").get(this.model.get("label").category.id);
 
                     title = modelJSON.label.abbreviation + " - " + modelJSON.label.value;
-                    this.$el.css("background-color", category.get("settings").color);
+                    var color = category
+                        ? category.get("settings").color
+                        : modelJSON.label.category.settings.color;
+                    this.$el.css("background-color", color);
                 } else {
                     title = modelJSON.text;
                 }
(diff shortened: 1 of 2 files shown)


frontend/js/views/list-annotation.js
```diff
@@ -352,9 +362,16 @@ define(["jquery",
                 modelJSON.duration     = (modelJSON.duration || 0.0);
                 modelJSON.textHeight   = this.$el.find("span.freetext").height();
 
-                if (modelJSON.isMine && this.scale && modelJSON.label.category.scale_id) {
+                if (modelJSON.label) {
                     category = annotationTool.video.get("categories").get(this.model.get("label").category.id);
 
+                    title = modelJSON.label.abbreviation + " - " + modelJSON.label.value;
+                    this.$el.css("background-color", category.get("settings").color);
```
Arnei, line 369: Does not handle undefined categories, which can for example happen if the category of the label was deleted.

JulianKniephoff: Oof, shoot. Why is that even allowed?!

JulianKniephoff: OK, I guess for now we have to deal with that. We should think hard about whether that's something we want, though. Maybe ... some time in the future ... A man can dream. Until then, I created #436.

I will update this shortly.

JulianKniephoff:
Note that this now uses #437, which should probably reviewed first.

Arnei (approved):
Approving this by virtue of being unable to break it.

PR #437, opened 2021-01-12
Add a method to easily get the correct color for an annotation

JulianKniephoff:
This will be used in #434 and #435.


Arnei (approved):
Approving this by virtue of it working in the other two PRs.

PR #439, opened 2021-01-18
Add information about docker-based solution to docs

hpawe01:


documentation/docker-environment.8.10.Dockerfile
```diff
@@ -0,0 +1,169 @@
+# Copyright 2016 The WWU eLectures Team All rights reserved.
```
mtneug, line 1: A simpler solution would be something like

```Dockerfile
FROM maven:3.6-jdk-8-slim AS build
RUN build maven ...

FROM quay.io/opencast/allinone:8.10
COPY --from build --chown=opencast:opencast /path/to/jars "${OPENCAST_HOME}/deploy"
RUN change config ...
```

This way upgrades of the underlying Opencast version are trivial and `Dockerfile` changes of the Opencast images do not need to be merged back.

mtneug (changes requested):

documentation/docker-environment.8.10.Dockerfile
hpawe01: This would be indeed much simpler, but I didn't try this approach, as in the [installation instructions](https://github.com/opencast/annotation-tool/blob/master/documentation/opencast-installation.md) of the annotation tool it is stated, that 
> You should make sure that your intended Opencast version was build at least once on the machine you use to build the Annotation tool

That is why I added the annotation tool build instructions within the step, where Opencast is built.

But I will try, if this is working.

mtneug: The Opencast JARs should be uploaded to Maven central and can be loaded from there. The problem is, that this process is a real pain and so sometimes this is not done right away. Therefore this recommendation since the JARs are then in your local Maven repository (~/.m2). However, I'm building the annotation tool now for a long time without a prepopulated local repository in a CI pipeline. This is definitely working with an older revision (you may want to enable the tests again):
```sh
mvn install --batch-mode \
  -Drevision="8" \
  -DskipTests=true \
  -Dcheckstyle.skip=true \
  -DskipJasmineTests=true \
  -DdeployTo="/path/to/deploy" \
  -Dopencast.version="[8,9-SNAPSHOT)"
```

lkiesow:
This is open and unfinished for a long time and does have conflicts. Let me dare to close this for now. But feel free to reopen this or create a new one when the problems are fixed.

PR #440, opened 2021-01-18
Fix deleting scales

JulianKniephoff:
Deleting a scale currently does not work in the sense that when you hit the delete button in the scale editor, nothing seems to happen. The console spews a few errors at you, though.

This **should** fix that.

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -1096,7 +1096,7 @@ define(["jquery",
                 },
                 destroy: function (scale, callback) {
                     _.invoke(
-                        _.clone(scale.get("scalevalues").models),
+                        _.clone(scale.get("scaleValues").models),
                         "destroy",
                         { error: function () { throw "cannot delete scale value"; } }
                     );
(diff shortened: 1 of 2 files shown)


Arnei (changes requested):
Seems to work, will still log a couple errors in the console. Here's for example what I got on firefox: 

`i18next::interpolator: failed parsing options string in nesting for key models.nominative SyntaxError: JSON.parse: unexpected character at line 1 column 14 of the JSON data
i18next.min.js:7:6922`

`Uncaught Error: undefined jquery-1.8.0.js:230:4
    valueOf self-hosted:2502`

Also not sure if this is related, but I cannot save edits to name and description of a scale .

JulianKniephoff:
I feel like this is the perfect analogy for the state of this project: I change one line, and my reviewer finds three new bugs while testing the one feature it touches. :angry: :disappointed: 

So okay, these indeed seem to mostly be unrelated: The `i18n` error should be fixed by #441, and the other errors that you saw (or at least the ones I saw, I hope we saw the same) were infinite recursion errors due to #442 which I don't yet have a good solution for. I think we just have to live with that one for now. :frowning_face: 

I'm not sure if I reproduced your lost edits issue correctly, but maybe #443 helps. :shrug:

Arnei (approved):
Seeing that all the errors I found are unrelated to this PR, I'm going ahead an approve this.

PR #443, opened 2021-01-19
Always start the scale editor in "not-edit mode"

JulianKniephoff:
Previously it already showed the editing interface for the scale
of the category you got to the dialog with. It wasn't fully in edit
mode yet, though, so edits could get lost.

### DIFF
--- a/frontend/js/views/scale-editor.js
+++ b/frontend/js/views/scale-editor.js
@@ -300,18 +300,9 @@ define(["jquery",
                     this.currentScaleId = this.$el.find("select#scale-id").val();
                     this.currentScale = annotationTool.video.get("scales").get(this.currentScaleId);
 
-                    if (this.currentScale && this.currentScale.isEditable()) {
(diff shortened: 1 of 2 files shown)


Arnei (commented):
Does solve the issue of changes being lost.

I'm not sure if I like this solution. If you now open the scale editor, you have to make an extra click to bring up the edit area, which feels somewhat pointless because there is no point in the edit area not being open. Furthermore, clicking "save" now closes the edit area, which also feels pointless . Either quit me out of the scale editor completely, or keep the edit area open.

JulianKniephoff:
It **is** weird. And you also have to do the extra click if you select another scale.

So I guess we would expect the edit area to just always be there, right? :thinking: Does this have any downsides?

I merged this into the Münster branch for now, because time pressure, but we don't have to merge it into `master`. I'll leave the PR open as a draft for now to remind me, since there is no issue for this.

Issue #442, opened 2021-01-19
You can't have two Bootstrap Modals open at the same time


JulianKniephoff:
Bootstrap modals try to keep your browser focus within themselves. This is a good thing, and we probably don't enforce this enough (looking at keyboard shortcuts for example), because this can lead to a multitude of race conditions we would need to handle if we don't assume this behavior.

However, if you have two modals, they currently just fight for the focus, leading to infinite recursion and thus to stack overflows.

I don't really know how we want to fix this, yet. :thinking: But it's a problem, for example, when you want to delete scale values or scales, since the delete modal is then issued from the modal scale editor.

JulianKniephoff:
We **might** get to this in the context of #455, but it's not high priority. :shrug:

PR #444, opened 2021-01-19
Fix double modals for scale(-value) deletion

JulianKniephoff:
Actually one solution to #442, just hiding one modal as long as another is open, was relatively easy to implement in the particular case of the delete modal and the scale editor being open at the same time.

Note that this includes #440, since that's the only way to even observe the faulty behavior right now.

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -67,7 +67,7 @@ define(["jquery",
                  * @param {Object} target Element to be delete
                  * @param {TargetsType} type Type of the target to be deleted
                  */
-                start: function (target, type, callback) {
+                start: function (target, type, confirmCallback, closeCallback) {
 
                     if (!target.isEditable()) {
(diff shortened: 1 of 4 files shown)


JulianKniephoff:
As with #444, I already merged this into the Münster branch, so there is no rush in getting this into `master`. We can leave it as a draft for now until we come up with something better. However, I think a proper workaround for #442 would take quite some time, so I don't know when this will be. :thinking: 

I would say we merge this, @Arnei creates another issue for the problem described, and I'll take a look at it soon-ish.

For Münster, what you describe is not a problem, because I had to refactor deleting scales (and scale values, labels and categories ...) quite a bit, and that seemed to have fixed that particular bug. And maybe that's even the solution for `master`: Just get the Münster changes merged and the problem goes away.

Arnei:
I agree with merging this, given that this is something that should be working already. 

If your work for Münster helps with fixing the other issue and can be integrated into master, all the better :)

PR #387, opened 2020-05-19
Frontend changes to enable multi content annotations

luniki:
This pull request will contain all changes required to enable multi content annotations.

What does this necessitate:

* [X] Change the data model
* [X] Add a way to add an empty multi content annotation
* [x] Change the way annotations are shown in the list of annotations
* [x] Enable adding new content to a multi content annotation
* [x] Enable dropping content from a multi content annotation
* [x] Enable editing content of a multi content annotation
* [x] Change the way annotations are shown in the timeline
* [x] Refactor existing code to use the new data structuring

Refs #370

PR #449, opened 2021-01-20
Multi Content Annotations

JulianKniephoff:
This supercedes #387.

#### Organizational

* [ ] This includes #443, which we might not want in `master` in it's current shape. A proper solution is needed here.

#### QA

* [ ] This needs another round of reviews. A lot of this code was written under time pressure and/or with incomplete information. We won't get it perfect, but I want to at least see if we can't throw out some of the more grievous smells.
  * [ ] Note that #387 already had a [review](https://github.com/opencast/annotation-tool/pull/387#pullrequestreview-434719236), and some of the comments remain unaddressed. You should incorporate these here.
* [ ] This PR currently fails CI because of test failures. These stem from changes in the API which the tests haven't been adapted to.

#### Bugs

* [ ] The CSV export still works, but doesn't understand MCA, yet, so it outputs mostly garbage.
* [ ] The Excel export and the print view don't work at all anymore.

#### (Missing) Features

* [ ] Questionnaires
  * [ ] Manage questionnaires per Opencast even/series
  * [ ] Questionnaire creator


JulianKniephoff:
Closing again. How all of this is going to make it's way back into `master` is not clear, yet. I pushed everything to the `questionnaire` branch for now.

PR #467, opened 2021-03-03
Fixed NaN widths for columns in the xlsx export

### LINKED ISSUE #466: XLSX-Export: All rows containing informations are hidden
If a user/admin does an export for Excel and opens it with Excel all rows containing informations are hidden. The column shown all left is the column "X". This leads to the perception, that the export did not work. In order to be able to see the export, the user first has to actively "Show" the hidden rows. See screencast: https://tube.switch.ch/videos/282116e3

Appears both ond Mac and Windows.

Arnei:
This PR fixes NaN widths for columns in xlsx files, which would cause Excel to hide the columns.

This was occuring for columns containing numbers, and in the case of a single row.

Resolves #466.

### DIFF
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -731,16 +731,13 @@ define(["jquery",
 
                 bookData.forEach(function (arr) {
                     Object.keys(arr).forEach(function (key) {
-                        let value = arr[key] === null ? '' : arr[key];
+                        var value = arr[key] === null ? '' : arr[key];
 
-                        let len = 0;
-                        switch(typeof value) {
-                          case "number": len = 10; break;
-                          case "string": len = value.length; break;
-                          case "object": if(value instanceof Date) len = 10; break;
-                        }
+                        // Arbitrarily increase len by one to avoid cases where just len would
+                        // lead to too small columns
+                        var len = value.toString().length + 1
 
-                        objectMaxLength[key] = Math.max(objectMaxLength[key] ? objectMaxLength[key] : 0, len);
+                        objectMaxLength[key] = Math.max(objectMaxLength[key] || 0, len);
                     });
                 });
 
--- a/frontend/js/annotation-tool.js
+++ b/frontend/js/annotation-tool.js
@@ -731,9 +731,16 @@ define(["jquery",
 
                 bookData.forEach(function (arr) {
                     Object.keys(arr).forEach(function (key) {
-                        var value = arr[key] === null ? '' : arr[key];
+                        let value = arr[key] === null ? '' : arr[key];
 
-                        objectMaxLength[key] = Math.max(objectMaxLength[key], value.length);
+                        let len = 0;
+                        switch(typeof value) {
+                          case "number": len = 10; break;
+                          case "string": len = value.length; break;
+                          case "object": if(value instanceof Date) len = 10; break;
+                        }
+
+                        objectMaxLength[key] = Math.max(objectMaxLength[key] ? objectMaxLength[key] : 0, len);
                     });
                 });
 


frontend/js/annotation-tool.js
```diff
@@ -731,9 +731,16 @@ define(["jquery",
 
                 bookData.forEach(function (arr) {
                     Object.keys(arr).forEach(function (key) {
-                        var value = arr[key] === null ? '' : arr[key];
+                        let value = arr[key] === null ? '' : arr[key];
```
JulianKniephoff, line 734:
```suggestion
                        var value = arr[key] === null ? '' : arr[key];
```

I know you like `let`, and I do agree with you (well, kind of; this variable could actually be `const`, even), but the consistent thing to do in this code base is unfortunately still `var`.

Some day we will get there, I hope ... :wink:

frontend/js/annotation-tool.js
```diff
@@ -731,9 +731,16 @@ define(["jquery",
...
 
-                        objectMaxLength[key] = Math.max(objectMaxLength[key], value.length);
+                        let len = 0;
```
JulianKniephoff, line 736:
```suggestion
                        var len = 0;
```

frontend/js/annotation-tool.js
```diff
@@ -731,9 +731,16 @@ define(["jquery",
...
+                        switch(typeof value) {
+                          case "number": len = 10; break;
+                          case "string": len = value.length; break;
+                          case "object": if(value instanceof Date) len = 10; break;
+                        }
+
+                        objectMaxLength[key] = Math.max(objectMaxLength[key] ? objectMaxLength[key] : 0, len);
```
JulianKniephoff, line 743:
```suggestion
                        objectMaxLength[key] = Math.max(objectMaxLength[key] || 0, len);
```

makes that line a bit shorter.

frontend/js/annotation-tool.js
```diff
@@ -731,9 +731,16 @@ define(["jquery",
...
-                        var value = arr[key] === null ? '' : arr[key];
+                        let value = arr[key] === null ? '' : arr[key];
 
-                        objectMaxLength[key] = Math.max(objectMaxLength[key], value.length);
+                        let len = 0;
+                        switch(typeof value) {
+                          case "number": len = 10; break;
+                          case "string": len = value.length; break;
+                          case "object": if(value instanceof Date) len = 10; break;
+                        }
```
JulianKniephoff, line 741:
```suggestion
                        switch (typeof value) {
                        case "number":
                            len = 10;
                            break;
                        case "string":
                            len = value.length;
                            break;
                        case "object":
                            if (value instanceof Date) len = 10;
                            break;
                        }
```

Some formatting here, but also can I ask how you came up with these cases? This seems a bit arbitrary. :thinking: And btw. what happens in the `"object"` case when it's not a `Date`?

And I might be wrong here, but the library probably puts something like `value.toString()` into the resulting file, right? So can't we just use `value.toString().length`? :thinking:

JulianKniephoff (changes requested):
Just some nitpicks, I guess. :sweat_smile:

frontend/js/annotation-tool.js
Arnei: Indeed very arbitrary, due to me not wanting to deal with proper column width calculation. 

Unfortunately just using `value.toString()` can often be insufficient. E.g. for the ID you would get
Bildschirmfoto vom 2021-03-24 11-11-36
instead of 
Bildschirmfoto vom 2021-03-24 11-15-59

I updated the PR with a different len calculation based on your suggestion. It is still arbitrary, but at least it is more concise. Please check if it is more to your liking.

JulianKniephoff: I feared it wouldn't be that trivial. :thinking: 

Alright, I made #474 to track the issue, but it's not a priority, and also not in the scope of this "hotfix" PR. This is good enough for now. :slightly_smiling_face:

PR #471, opened 2021-03-11
Fixes the overflow behavior of the annotate container

### LINKED ISSUE #463: Fix the overflow behavior of the annotate container

marwyg:
Fixes #463

While rescaling the annotation container in edit mode, the MCA button on the bottom should stay visible. Also the vertical and horizontal scrollbars should appear and stay visible while rescaling.

### DIFF
--- a/frontend/style/bootstrap/carousel.less
+++ b/frontend/style/bootstrap/carousel.less
@@ -4,6 +4,8 @@
 
 
 .carousel {
+  height: 0;
+  flex-grow: 1;
   position: relative;
   margin-bottom: @baseLineHeight;
   line-height: 1;


JulianKniephoff (changes requested):
This seems to be the right approach. However, the place where you made these changes is kind of off-limits. Unfortunately the annotation tool doesn't use a package manager to pull in its dependencies (yet!). Instead, they are just pasted into the source tree. This file is part of such a "vendored" package (Bootstrap), which we shouldn't change. (That way lies madness.)

Instead, I would suggest finding an appropriate place somewhere in `frontend/style/annotations/annotate.less`. The folder `frontend/style/annotations` contains all the custom styles we use, somewhat sorted by component. Everything having to do with creating annotations is in `annotate.less`. The other subfolders of `frontend/style` all belong to some package and should not be changed except for being deleted because we don't need it anymore, or being updated wholesale with a new version of the package.

PR #472, opened 2021-03-11
Imported labels doesn't disappear anymore

### LINKED ISSUE #464: Imported labels disappear until the next reload

marwyg:
Fixes #464 

After importing annotations, the labels should stay visible as expected.

### DIFF
--- a/frontend/js/views/annotate-category.js
+++ b/frontend/js/views/annotate-category.js
@@ -287,6 +287,11 @@ define(["jquery",
             removeOne: function (delLabel) {
                 _.find(this.labelViews, function (labelView, index) {
                     if (delLabel === labelView.model) {
+                        // this event fires when the "deleted_at" attribute of the label model changes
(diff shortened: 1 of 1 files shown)


JulianKniephoff (changes requested):
Right, that makes sense. >_< Thanks for finding the issue.

However, I would do it slightly differently. GitHub's review feature wasn't sufficient to suggest it here, so I made you a PR over at your fork: https://github.com/marwyg/annotation-tool/pull/1

Are you okay with that? :slightly_smiling_face:

JulianKniephoff:
We are currently trying to clean up this repository a bit in preparation for a community-wide discussion about the future of the project. Since this still has some work to do, but we currently lack the capacity to follow up on it, I'll close this for now. We can always bring it back if/when we take up the questionnaire changes again.

JulianKniephoff:
For potential future reference, see also https://github.com/marwyg/annotation-tool/pull/1.

Issue #454, opened 2021-02-04
Option to pause the video on structured annotations


JulianKniephoff:
This is currently only possible for free text annotations, where you can check a checkbox and have the tool pause the video once you start typing in the appropriate text field. It also automatically resumes when you eventually create the annotation.

JulianKniephoff:
@dagraf I'm not quite sure I understand this correctly. For structured annotations you just want the video paused the moment someone clicks on a label and/or scale value? Should it also auto-resume at some point? When?

dagraf:
@JulianKniephoff Yes, correctly understood. I see no smart solution for an auto-resume. Therefore I would suggest not to implement one. Do you have a suggestion? But please make sure, that hitting the space bar does resume the video.

JulianKniephoff:
No, I don't see a good way to add auto-resume, either. :thinking: 

Just out of curiosity: What exactly is the use case for this feature? That is, what do you expect a user to do once the video is paused after inserting a structured annotation?

dagraf:
Main use case: User has the task to mark a sequence with a structured annotation and to add some thoughts using the comment-function or by writing a freetext annotation at the exact same time.

JulianKniephoff:
OK, thanks, that makes sense. :slightly_smiling_face:

dagraf:
Specification: A click in textbox should NOT stop video. 

Reasoning: 
- if I write a comment, insert it by hitting with enter and click back into the textbox, the video does not stop. But if I insert a comment by clicking on the "Insert" button and then click back into the commentbox, the video does stop. This different behavior can confuse users.
- The description is called differently ("Paus video while writing an freetext annotation").
- The productive tool does work as described above (video does not stop when clicking in the textbox).

PR #475, opened 2021-03-25
Add an option to pause the video after creating a structured annotation

JulianKniephoff:


JulianKniephoff:
The checkbox labels aren't final, yet. Waiting for @dagraf. I will deploy this to staging by next week, hopefully.

Arnei (commented):
Does not seem to work with scales (when creating a labelled annotation with a scale). Not sure if this is intended or not.

Also the current UI seems to waste a bit of space. Might be nice to make it more flexible. Might also be confusing though.
Bildschirmfoto vom 2021-04-08 10-56-33
editIdea

JulianKniephoff:
> Does not seem to work with scales (when creating a labelled annotation with a scale). Not sure if this is intended or not.

Nope, it's not, good catch. Should be fixed now.

> Also the current UI seems to waste a bit of space. Might be nice to make it more flexible. Might also be confusing though.

I see Pros and Cons for both, with a slight preference for the "vertical layout." @dagraf opinions?

Arnei:
> Nope, it's not, good catch. Should be fixed now.

Seems to be fixed.

> I see pros and cons for both, with a slight preference for the "vertical layout." @dagraf opinions?

I was imagining some flexbox thing where the layout changes based on the available width. Although I'd absolutely agree that a vertical layout has better visual clarity (as that is what I would expect out of a checkbox list). Wouldn't want this to hinder the merge tho.

Also, unrelated to the this PR, but what counts as "while editing free text annotations", meaning in which cases the video automatically resumes play after pausing due to a free text annotation "edit", seems weird to me (Click in the category box? Resume play. Click on the timeline? Do not resume play. Open a different window altogether? Resume play). Might be worth a new issue.

dagraf:
**Label of checkboxes:** I do not really like the term "structured annotation" because it does not exist elsewhere in the annotation tool. I suggest we use the same wording we are planning to use for the upcoming modals:
- DE: "Video nach Klick auf eine Label / Etikett anhalten"
- EN: "Pause vide after clicking on a label"

English text "while editing free text annotations": I think this is due to a faulty translation of the german "Video während des Schreibens anhalten". So I suggest to alter it to: "Pause video while writing free text annotations"

dagraf:
Position of new option: I would prefer the vertical layout. But in fact, we would suggest do place it right beneath the box for the structured annotations. So we have some logic: first the box for writing or selecting an annotation and right below the option to pause the video when I make an annotation (see mockup). 

Bildschirmfoto 2021-04-12 um 14 40 24

Please pay attention to the additional grey space we entered underneath the existing checkbox "pause while editing free text annotations" and the three tabs.

dagraf:
Additionally, I would suggest that we consider to change the default setting: 
- for free text annotations, the checkbox is by default activated.
- for structured annotations, the checkbox is by default DEactivated.

dagraf:
Other possibility for where to put and how to name the two pause options: 
Bildschirmfoto 2021-04-12 um 17 23 24

JulianKniephoff:
@Arnei raises a good point, though: We need to think about how all these elements behave when we shrink that view. But I think that's a question for another day as well.

As for the wording and placement: I don't really have any strong opinion. If @Arnei doesn't have any objections I would just say: @dagraf pick your favorite and that's what we'll go with. :)

JulianKniephoff:
> Also, unrelated to the this PR, but what counts as "while editing free text annotations", meaning in which cases the video automatically resumes play after pausing due to a free text annotation "edit", seems weird to me (Click in the category box? Resume play. Click on the timeline? Do not resume play. Open a different window altogether? Resume play). Might be worth a new issue.

Looking at the code, the rules are apparently:

* Resume whenever I submit the current free text annotation
* Resume ~200ms after the input field loses focus

I don't know how reasonable of a definition that is from a user's point of view. If not, we should indeed have an issue for that. @dagraf care to chime in?

If that is in fact what we want, but we are seeing behavior that differs from that, I would also welcome an issue. :)

dagraf:
> As for the wording and placement: I don't really have any strong opinion. If @Arnei doesn't have any objections I would just say: @dagraf pick your favorite and that's what we'll go with. :)

My favorite:
- Wording: "Video anhalten:" / "Während des Schreibens einer Freitextannotation" / "Beim Setzen einer strukturierten Annotation"
- Placement: Below the box for structured annotations.

JulianKniephoff:
The problem with the interaction between the two options I mentioned above should be fixed now. I'll use this opportunity to think through some more edge cases and maybe make the system a bit more robust, still.

dagraf:
> > Also, unrelated to the this PR, but what counts as "while editing free text annotations", meaning in which cases the video automatically resumes play after pausing due to a free text annotation "edit", seems weird to me (Click in the category box? Resume play. Click on the timeline? Do not resume play. Open a different window altogether? Resume play). Might be worth a new issue.
> 
> Looking at the code, the rules are apparently:
> 
> * Resume whenever I submit the current free text annotation
> * Resume ~200ms after the input field loses focus
> * If the video seeks for any reason before that, it will **not** resume
> 
> I don't know how reasonable of a definition that is from a user's point of view. If not, we should indeed have an issue for that. @dagraf care to chime in?

I'd try to take the user's point of view when to resume the video: 
- When the video got paused because of writing a freetext annotation: (1) After I insert this annotation I was writing (ether by hitting 'Enter' or clicking on 'Insert', (2) After clicking on the play button in the viewer
- When the video got paused because of setting a structured annotation: (1) After clicking on the play button in the viewer

JulianKniephoff:
@dagraf So specifically if you just leave the textbox **without** submitting, you would **not** expect the video to resume, right? Because I was certainly surprised when I just saw that in the code. x)

dagraf:
> @dagraf So specifically if you just leave the textbox **without** submitting, you would **not** expect the video to resume, right? Because I was certainly surprised when I just saw that in the code. x)

Correct.

Arnei:
> As for the wording and placement: I don't really have any strong opinion. If @Arnei doesn't have any objections I would just say: @dagraf pick your favorite and that's what we'll go with. :)

Fine by me, I like @dagraf 's suggestions.

JulianKniephoff:
@dagraf one more question to clarify: Currently, when the video pauses because of the "pause on freetext" option, and then you jump to another position in the video, the video will **not** resume after submitting the annotation. Is this correct?

JulianKniephoff:
Uh, and another one. :sweat_smile: 

What should happen if I uncheck the option before submitting my annotation? Should it resume?

dagraf:
> @dagraf one more question to clarify: Currently, when the video pauses because of the "pause on freetext" option, and then you jump to another position in the video, the video will **not** resume after submitting the annotation. Is this ~correct~ what you want?

No, this is not what I want or what I would expect. I would expect that after submitting the annotation the video does resume.

dagraf:
> Uh, and another one. 😅
> 
> What should happen if I uncheck the option before submitting my annotation? Should it resume?

Yes, it should resume. But either way would be ok. Because I guess this would happen rather rarely.

JulianKniephoff:
> Yes, it should resume. But either way would be ok. Because I guess this would happen rather rarely.

Ok, that's good to hear, because thinking about it some more, this really seems like the only option. This is a surprisingly tricky feature, mainly because the tool doesn't really have a discrete notion of the event of creating a free text annotation. :thinking:

Anyway, I think I got it now. :slightly_smiling_face:

JulianKniephoff:
OK so quick ad hoc TODO list for me so I don't forget anything.

* [x] change the behavior of the "pause for freetext" option to what we just discussed
* [ ] Change the wording and placement of the options
* [ ] Change the defaults of the options

Am I missing something?

JulianKniephoff:
Alright, this should be ready for another review. @dagraf I might throw this and #478 onto staging as well, later today, so you can maybe test it tomorrow or on Monday before our meeting. :smiley:

dagraf:
That would be great. Please let me know as soon as I can test it on staging.

JulianKniephoff:
@dagraf I deployed the two patches; however, I can't test #478 myself right now, I think, because our test user that should have sufficient permissions has expired.

dagraf:
Which test user are you working with?

JulianKniephoff:
`stp-OrgAdmin1` should have the correct rights, which is what I was trying. `stp-Producer1` does not seem to have them, although maybe that's already a bug. :thinking:

JulianKniephoff:
I just now got a mail that the user works again, and I still don't see the new visibility option. That's weird. I will look into it ASAP.

JulianKniephoff:
@dagraf if I remember correctly you wanted more spacing around the options. How about this?

image

JulianKniephoff:
The thing you mentioned in https://github.com/opencast/annotation-tool/issues/454#issuecomment-830221212 is also changed. Will be effective the next time I deploy.

dagraf:
> @dagraf if I remember correctly you wanted more spacing around the options. How about this?
> 
> image

Correct. Looks fine. Please make shure that the spacing above the pause video option is a bit smaller than the spacing below it. Thx!

Arnei (approved):

JulianKniephoff:
> Please make shure that the spacing above the pause video option is a bit smaller than the spacing below it. Thx!

Like this?

image

dagraf:
Yes. thx!

PR #478, opened 2021-04-19
Example tracks

### LINKED ISSUE #374: Track example for students
Sometimes instructors create tracks with example annotations to show students when to use which annotations. Instructors should be able to specify tracks which are always visible upon loading the annotation tool.

JulianKniephoff:
This will close #374.

Currently this does not allow any write operations by other users.


Arnei (approved):
Works.

I do find the naming of the new option somewhat unintuitive, given that it is the only option that shoves the track in everybody's faces. I've yet to come up with a naming that satisfies me though ("Forcibly display to all users" does not sound very nice after all).

dagraf:
I agree with @Arnei. What about: "Automatically display to all users"?

JulianKniephoff:
Rebased onto `master` after merging #477.

JulianKniephoff:
I'm also not satisfied with the wording. :thinking: 

What @dagraf suggested seems reasonable, though. @Arnei objections?

Arnei:
Seems good.

JulianKniephoff:
This is deployed on SWITCH staging right now, but none of my test users can see the new visibility option. (Only annotate admins are supposed to see those.) This might be due to missing ACL entries for these users, or it might be a bug. Will investigate ASAP.

JulianKniephoff:
Indeed seems to be a problem of the user; `stp-OrgAdmin1` is surprisingly not an Org-Admin. :sweat_smile:

PR #481, opened 2021-04-27
Clean up the frontend build system some

### LINKED ISSUE #483: Fix source maps for the `less` files

JulianKniephoff:
Big parts of the `Gruntfile` are currently rarely or never exercised. They are likely broken, or will be in the foreseeable future.

This PR removes a bunch of it, and also all the dependencies it brought, closing our last (currently known) security vulnerability.

Specifically, this now also removes the `grunt dev`-setup. To test the Annotation Tool during development, you now always need a running Opencast instance to install it into. I thought a lot about this, but I want to get rid of all the `localStorage`-specific stuff in the code, that is basically just there to support his use case. I have other ideas to make the feedback cycle during development quicker.

Incidentally this now also closes #483.

~~There might be more to come, but what's here is already ready for review and should already be an improvement over the status quo, so depending on whether I or @Arnei is faster, I will amend this PR or make a new one for the rest. :wink:~~ Ok, there is of course always more one can do, but I think this is a pretty good PR now. I should now go do other things again. :sweat_smile:


Arnei (approved):
I honestly don't a lot about the major things that got removed here (e.g. grunt), but the changes looks fine and nothing seems to be broken when I run this in my local Opencast, so I'll approve this.

Couple questions:
1. Are we aiming to completely get rid of grunt and the associated way of building the frontend seperately? Does this mean we do not want to provide a way to build the Annotation-Tool as a standalone frontend anymore?
2. Why do you dislike the comments in the `Gruntfile`? They don't seem to be wrong.

JulianKniephoff:
> 1. Are we aiming to completely get rid of grunt and the associated way of building the frontend seperately? Does this mean we do not want to provide a way to build the Annotation-Tool as a standalone frontend anymore?

My dream would be to replace Grunt with something more modern like Webpack, but who knows if that will ever happen. There will however always be **some** kind of separate (from Maven) build system for the frontend; I don't want Maven to do it all. I would actually like to extend the capability to build the tool separately as in not necessarily for the purpose of being bundled into a JAR.

However, the capability to **run** separately, i.e. without the backend, using `localStorage` is something that I want to get rid of in the long run, yes. I thought long and hard about this, and it makes me a bit sad, but the current implementation only makes the code more complex in many places without serving any real need. No one (to my knowledge) uses the tool like that. Well, I used it like that for development quite often because it was quicker than to whip up an Opencast. But I have other plans to make the development process more comfortable.

> 2. Why do you dislike the comments in the `Gruntfile`? They don't seem to be wrong.

Not all of them were wrong, granted, but most of those that weren't were redundant in my opinion. Things like:

```js
/* Copy ... */
copy: {
    ...
}
```

just add noise in my opinion. Maybe I go overboard with this attitude sometimes, but especially with the `Gruntfile` being reduced as much as it is now, I felt that they didn't really serve any function anymore.

PR #486, opened 2021-05-06
Remove `localStorage` specific code

JulianKniephoff:
Some follow up to #481. Since the "`dev`-mode" is no longer supported, we can get rid of code that was purely there to support it.


Arnei (approved):
Does not seem to break anything and less dead code is great. (Also the "scalevalue/scaleValue" part is finally gone, hooray)

PR #491, opened 2021-05-07
Refactor the way resources fetch their children

JulianKniephoff:
Specifically this gets rid of our `validate` heck
where we check for changes to the `id` field.


Arnei (approved):
Way more concise, lovely. Also does not seem to break on exisiting events.

PR #494, opened 2021-05-17
Clean up the source tree a bit

JulianKniephoff:
This mainly gets rid of the weird `build` folder and moves its contents to appropriate places within `js`.

As usual I used the opportunity to fix some minor things I stumbled upon as well, though.


Arnei (approved):
Builds and runs.

PR #495, opened 2021-05-17
Put a basic linting setup in place

JulianKniephoff:
This really doesn't enforce much, yet. The major exception is that it parses the main `js/**/*.js` files as "pre-ES5". This (for now) precludes the use of `let`, etc. (Sorry @Arnei. I promise we will get there some day. Having a linter is good first step in that direction. :wink:)

This is based on #494, so that should probably be reviewed/merged first.


Arnei (approved):
Hooray for linting. Works (I am indeed not allowed to use `let` anymore) and builds.

PR #499, opened 2021-06-02
Improve the project structure slightly

JulianKniephoff:
The main thing this PR does is getting rid of the artificial `frontend-delivery` project/bundle/thing. The `frontend/pom.xml` now directly bundles the results of the Grunt build.

As usual, some minor nitpicks I found while doing this are included.

Note that this makes it a bit simpler to hack just the frontend again. Once you have an Opencast instance with a running Annotation Tool, you merely need to rebuild and redeploy the `frontend` bundle, instead of building `frontend` and then building and deploying `frontend-delivery` (or just building and deploying the entire project again, which is more convenient but also takes longer).


Arnei:
There are a couple of "TODO" comments in there, are those intended?

JulianKniephoff:
Whops. Nope, those are just for me to ponder another time. Should be gone now. :sweat_smile:

Arnei (approved):
Works in my local installation and seems reasonable as far as I can tell

JulianKniephoff:
Fixed CI; this should not change anything about @Arnei's review, so I'm merging this now.

PR #501, opened 2021-06-25
Fix the backend tests

JulianKniephoff:
The `contentType` calls are apparently necessary because
when you don't specify a `formParam` in a call to REST Assured,
it defaults to a `ContentType` of something other than
`application/x-www-form-urlencoded`, which the JAX-RS stuff
refuses to answer in the presence of `@FormParameter` annotations.

This was different when we still mostly dependet/ran against
Opencast 8, so I'm thinking something in the REST test environment
we use changed. ¯\\\_(ツ)_/¯

---

Closes #325

PR #502, opened 2021-06-29
Update to Opencast 9/10

JulianKniephoff:
This also requires us to migrate away from
the `SimpleServicePublisher`.
I used that opportunity to convert the project to using
OSGi annotations instead of the more error prone XML based
service declarations.

---

**Note**: This includes #501 so that I could test this stuff effectively.


Arnei:
I'm running into the below error when trying to load an event with the annotation tool.
```
XHR PUT http://localhost:8080/extended-annotations/users [HTTP/1.1 405 Method Not Allowed 0ms]
```
Running Opencast 10 on CentOS. I'm fairly certain I set up permissions correctly.

JulianKniephoff:
Do you see anything interesting in the logs? :thinking: Can you just send me everything that's in the logs after you installed the tool?

Arnei:
Tried again and it works now. I suppose I made some mistake yesterday.

JulianKniephoff:
Yeah no, it's weird. I had something similar happen at first, but then it magically worked. I noticed this behavior a few times before, too, though, when switching Opencast versions or branches or something like that. I **still** don't know what causes this. :shrug:

PR #516, opened 2021-09-09
Fix category clean-up when tracks are not visible

Arnei:


JulianKniephoff (approved):
Oh, whoops. :sweat_smile:

PR #544, opened 2021-10-28
Some cleanup

JulianKniephoff:
Some code quality things I came up with during my work on #478. Unfortunately these never made it to SWITCH staging, so they could use a second pair of eyes before I just merge them into `master`. :wink:


frontend/js/views/main.js
```diff
@@ -814,10 +814,6 @@ define(
                 this.loadingBox.find(".bar").width(this.loadingPercent + "%");
                 this.loadingBox.find(".info").text(message);
             }
-        }, {
-            EVENTS: {
-                READY: "ready"
-            }
```
Arnei, line 815: Why do you not like the constant/ why do you think it's redundant?

frontend/js/models/video.js
```diff
@@ -100,35 +100,39 @@ define(
...
-                var result = [];
-                var handleAnnotation;
-                if (_.isUndefined(category)) {
-                    handleAnnotation = _.bind(Array.prototype.push, result);
-                } else if (category) {
-                    handleAnnotation = function (annotation) {
+            getAnnotations: (function () {
+                return function (category) {
+                    return this.get("tracks").chain()
+                    // TODO This needs Underscore 1.9
```
Arnei, line 110: Looks like this TODO is done

Arnei (approved):

frontend/js/views/main.js
JulianKniephoff: What bugs me most is that it is not used consistently, and since not using the constant is conceptually the simpler "tactic" it's what I would move towards by default.

In this particular case I would add, though, that while constants can of course be useful I'm suspicious of those of the form:

```
const FOO = "foo";
```

They don't really offer any value. Sure, it "hides" the magic value `"foo"`, but if that ever changes, are you really going to live with a code base that has

```
const FOO = "bar";
```

in it? I wouldn't want that. :thinking: So you still end up having to change it everywhere. For that purpose `MainView.EVENTS.READY` might still be easier to find than `"ready"`, granted, but this is JavaScript; you have to be careful anyway, since I can also refer to that constants through many different names.

PR #545, opened 2021-11-01
Update to a more current Node.js and NPM version

JulianKniephoff:
Note that if you test this in a "used" working directory, you  might have to run `rm -rf frontend/node*` for this to build.


Arnei (approved):
Maybe that note should be added to the docs as a troubleshooting tip. Ala "If your build does not work for whatever reason, run this command and try again".

PR #546, opened 2021-11-01
Create a nice base for modal views and refactor some of our code to use it

JulianKniephoff:
We have quite a few modals by now, and a few more are coming as part of Bern 2021-1. This provides a nice base with some of the boilerplate code necessary to make a modal view. It also refactors some of our code to use it, though not all of it unfortunately. What's missing is the delete modals which are ... special, and the scale editor which will be rewritten as part of Bern 2021-1.

### DIFF
--- a/frontend/js/views/alert.js
+++ b/frontend/js/views/alert.js
@@ -19,16 +19,14 @@
  */
 define(
     [
-        "jquery",
         "underscore",
-        "backbone",
+        "views/modal",
         "templates/alert-modal",
         "bootstrap"
     ],
     function (
-        $,
         _,
-        Backbone,
+        Modal,
         AlertTemplate
     ) {
         "use strict";
@@ -39,10 +37,7 @@ define(
          * @augments module:Backbone.View
          * @memberOf module:views-alert
          */
-        var Alert = Backbone.View.extend({
-
-            el: $("#alert"),
-
+        var Alert = Modal.extend({
             /**
              * Alert template
              * @type {HandlebarsTemplate}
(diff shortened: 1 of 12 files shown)


frontend/locales/de/translation.json
```diff
@@ -58,24 +57,18 @@
     "track": "auf",
     "comments": {
       "author": "Kommentar von",
-      "cancel": "$t(common actions.cancel)",
       "count": "Ein Kommentar",
       "count_plural": "{{count}} Kommentare",
       "created": "am {{date}}",
-      "delete": "$t(common actions.delete)",
       "edit": {
         "button": "Bearbeiten",
```
Arnei, line 64:
```suggestion
        "button": "Bearbeiten"
```

frontend/locales/de/translation.json
```diff
@@ -300,20 +285,51 @@
...
+    },
+    "scale value": {
+      "identifier": "Skalenwert",
+      "gender": "male"
+    },
+    "scale": {
+      "identifier": "Skala",
+      "gender": "female"
+    },
+    "determined": "$t(determiners.{{determiner}}.{{case}}.$t(models.{{context}}.gender))",
```
Arnei, line 316:
```suggestion
    "determined": "$t(determiners.{{determiner}}.{{case}}.$t(models.{{context}}.gender))"
```

frontend/locales/de/translation.json
```diff
@@ -300,20 +285,51 @@
...
+        "male": "den",
+        "female": "die",
+        "neutral": "das"
+      }
+    },
+    "demonstratives": {
+      "accusative": {
+        "male accusative": "diesen",
+        "female accusative": "diese",
+        "neutral accusative": "dieses",
```
Arnei, line 330:
```suggestion
        "neutral accusative": "dieses"
```

frontend/locales/de/translation.json
```diff
@@ -300,20 +285,51 @@
...
+      }
+    },
```
Arnei, line 332:
```suggestion
    }
```

Arnei (changes requested):
Trailing commas are preventing the german translation from appearing at all.

Arnei (changes requested):
The german translation for the delete modal Text "Wollen Sie [...] wirklich löschen?" does not work. Unfortunately it's somewhat hard for me to figure out how this is supposed to work, so I don't have fix suggestion.

frontend/js/views/alert.js
```diff
@@ -105,26 +80,27 @@ define(
             TYPES: {
                 ERROR: {
                     title: "alert.error.title",
-                    className: "alert-error"
+                    severity: "alert-error"
                 },
                 FATAL: {
                     title: "alert.fatal.title",
                     className: "alert-error",
```
Arnei, line 87: Should probably also be:
```suggestion
                    severity: "alert-error",
```

frontend/js/views/tracks-selection.js
```diff
@@ -92,20 +97,14 @@ define(
...
              */
             initialize: function () {
-                _.bindAll(
-                    this,
-                    "show",
-                    "hide",
-                    "search"
-                );
+
+                Modal.prototype.initialize.apply(this, arguments);
```
Arnei, line 101: Comprehension questions: Should this really be called on the prototype? Wouldn't this apply the arguments to ALL `views/modal` modals? Might be me misunderstanding prototypes, but an explanation would be nice :D

JulianKniephoff:
> The german translation for the delete modal Text "Wollen Sie [...] wirklich löschen?" does not work. Unfortunately it's somewhat hard for me to figure out how this is supposed to work, so I don't have fix suggestion.

It took some trial and error and in the end some hacking, but I could fix it in 263e537.

PR #547, opened 2021-11-01
New way of editing categories and scales

JulianKniephoff:
This implements a new paradigm for editing categories and scales, as prescribed by the Bern 2021-1 project.

Note that this includes #546 and Arnei/annotation-tool#8, and hence Arnei/annotation-tool#7 and #414.


frontend/js/views/list-annotation.js
```diff
@@ -221,11 +221,11 @@ define(
...
-                var $target = $(event.currentTarget),
-                    value = $target.val(),
-                    radix = 10, // Radix is 10 for decimal
-                    values,
-                    seconds;
+                var $target = $(event.currentTarget);
+                var value = $target.val();
+                var radix = 10; // Radix is 10 for decimal
+                var values;
+                var seconds;
```
Arnei, line 228: Super minor thing: From the changes I've inferred that you like to declare variables where they're initialized whenever possible. This is the only place in the code where you broke that rule. Any reason for this?

frontend/templates/scale-modal/scale-editor.tmpl
```diff
@@ -0,0 +1,22 @@
+{{#unless isNew}}
```
Arnei, line 1: The delete button for scales is never shown, probably because this is always true?

frontend/templates/category-modal.tmpl
```diff
@@ -0,0 +1,89 @@
...
+            </label>
+            <label class="radio">
+                <input type="radio" name="affiliation" value="series">
+                {{t "category modal.belongs to series"}}
+            </label>
+
+            <span class="help-block" id="no-series-affiliation-with-scale">
+                {{t "category modal.no series affiliation with scale"}}
+            </span>
+        </fieldset>
```
Arnei, line 32: The "Belong to video or series"-radio button are currently always displayed. I personally think it would make more sense to allow users to only affiliate a category with a series when that event actually belongs to a series. 

Furthermore, this seems to break things when:
1. On an event that does not belong to a series, create a public category that belongs to a series.
2. Create some annotations with the category.
3. Change the event so that it belongs to a series.
4. The created annotations cannot be fetched anymore, and new ones cannot be created using this category.

Arnei (changes requested):
This has been less of a code review and more a "Arne learns about javascript" session, but I still daresay this PR looks generally good to me. Some issues I found are listed below.

Should this include a notice about upgrading the db, since a new a new column was added to LabelImpl?

The text in the "Add Track"-Modal is all white, making it unreadable against the white background. Seems to be an inheritance issue somehow.

frontend/js/models/scale.js
```diff
@@ -82,6 +80,18 @@ define(
...
             },
 
+            /**
+             * Parse the attribute list passed to the model
+             * @param {object} data Object literal containing the model attribute to parse.
+             * @return {object} The object literal with the list of parsed model attribute.
+             */
+            parse: function (data) {
+                if (data.scaleValues) {
+                    data.scaleValues = new ScaleValues(data.scaleValues, { scale: this });
```
Arnei, line 90: This seems sketchy somehow, might be worth explaining in a comment why this is necessary.

frontend/js/views/list-annotation.js
JulianKniephoff: As far as I can tell I didn't write that code; I only touched it in some linting changes. I could have gone further, sure.

So yeah, no, no particular reason.

JulianKniephoff:
> Should this include a notice about upgrading the db, since a new a new column was added to LabelImpl?

Unfortunately we still haven't figured out a good way/place to put such notices. :disappointed: 

> The text in the "Add Track"-Modal is all white, making it unreadable against the white background. Seems to be an inheritance issue somehow.

Will be addressed in #548.

frontend/templates/scale-modal/scale-editor.tmpl
JulianKniephoff: This will be addressed in #568 (once I resurrect it and port it to the appropriate branch).

frontend/templates/category-modal.tmpl
JulianKniephoff: I agree with the first part, but integration such logic would require some more work we don't have time for, currently. :disappointed: 

Similarly, either one of Lars' outstanding PRs will fix the bug you describe, OR we will accept it for now, until someone orders us to fix it. :sweat_smile:

Issue #533, opened 2021-10-12
Sharing option: Additional explanation for new option


dagraf:
When option "Show all users automatically" is actived, please show below the following message:
DE: Die Spur wird allen Benutzern automatisch und schreibgeschützt angezeigt. Andere Benutzer können also keine Änderungen vornehmen.
EN: The track is displayed to all users automatically and in read-only mode. So other users can not make any changes.

dagraf:
Text is now there. But because is it in the same format and font as the option itself it is not clear on the first sight, what is the option and what the explanation. Could you please format the explanation slightly differently? For example in grey, in italics or a bit intended?

dagraf:
Thx for your suggestion. I agree, it is worth thinking about how the naming of the access options could be optimised. But the naming you suggest doesn't really fit to what the options are doing. Let me explain what they do:

- Automatically display to all users: Makes a track public and takes care of showing this track to every user who opens the video in the annotation tool. But the track is still in read only mode for other users.
- Public: Makes a track public. But a user has to actively 'open' the track using the funcionality "Track management".

Do you have other ideas how to improve the naming?

lkiesow:
Are the tracks in read-only mode if I select “Public” or “Share with admin”? Or can anyone/admin then edit them as well?

dagraf:
Tracks are always in read-only mode.

lkiesow:
In that case, let me make a suggestion based on my current understanding:

- Global preview (automatically shown to all users)
- Public (any user may access this)
- Shared with admin (admins may access this)
- Private (only you have access)

My idea is foremost to make a distinction between what the new state does in contrast to public.
If you want, we could also highlight the names a bit more. Something like:

- __Global preview__ (automatically shown to all users)
- __Public__ (any user may access this)
- __Shared with admin__ (admins may access this)
- __Private__ (only you have access)

…and now you tell me what I did fail to understand this time :wink:

dagraf:
:) No misunderstandings anymore.
I like your suggestion including the highlighting of the names. But I am not quite sure about the name "Global preview", mainly because it is not e "preview" (according to leo.org: Vorschau, Probevorführung, ...). Here my suggestion: 

- **Public & by default** (automatically shown to all users)
- **Public & on demand** (any user may access this)
- **Shared with admin** (admins may access this)
- **Private** (only owner has access)

German:
- **Öffentlich & automatisch** (wird allen Usern automatisch angezeigt)
[...]

PR #548, opened 2021-11-08
Explain example track

JulianKniephoff:
This addresses #533.

Note that this includes #546 and #544.


frontend/locales/de/translation.json
```diff
@@ -222,7 +222,8 @@
       "to public": "Öffentlich machen",
       "to private": "Privat machen",
       "to shared with admin": "Mit Administratoren teilen",
-      "to shared with everyone": "Allen Benutzern automatisch anzeigen"
+      "to shared with everyone": "Allen Benutzern automatisch anzeigen",
+      "to shared with everyone explanation": "Die Spur wird allen Benutzern automatisch und schreibgeschützt angezeigt. Andere Benutzer können also keine Änderungen vornehmen."
```
Arnei, line 226: This line is in a different place than in the english translation file and therefore doesn't get shown. Apparently it should be under "timeline.access" (and without the "to").

Arnei (changes requested):
Also agree with @dagraf s comment in #533 it would be nice if the explanation could be formatted slightly differently. Looks good to me otherwise.

JulianKniephoff:
> Also agree with @dagraf s comment in #533 it would be nice if the explanation could be formatted slightly differently.

The explanation is going to change again anyway in #567.

PR #558, opened 2022-01-23
Fix UnusedImport

lkiesow:
This fixes a checkstyle issue which was complaining about an unused
import.


JulianKniephoff:
This is a change specific to the SWITCH version of the code. It is merged there and will land here if/when we get to backporting their customizations.

PR #559, opened 2022-01-23
Fix REST Docs exception

lkiesow:
This patch fixes the problem with Opencast's REST docs presenting an
exception when trying to view the endpoint documentation.

This does not address the completely missing documentation for all the
Annotation Tool endpoints.


opencast-backend/annotation-impl/src/main/java/org/opencast/annotation/endpoint/ExtendedAnnotationsRestService.java
```diff
@@ -34,6 +35,12 @@
         "opencast.service.type=org.opencast.annotation",
         "opencast.service.path=/extended-annotations"})
 @Path("/")
+@RestService(
+    name = "extended-annotations",
+    title = "Annotation Tool Backend",
+    abstractText = "Scientific video annotations on Opencast media..",
+    notes = { }
```
lkiesow, line 42:
```suggestion
    notes = { "The Annotation Tool does not yet provide REST documentation." }
```

PR #560, opened 2022-01-23
Prevent NullPointerException

lkiesow:
This patch prevents a NullPointerException whenever someone would access
the Annotation Tool with a media package identifier unknown to Opencast.


JulianKniephoff:
This is a change specific to the SWITCH version of the code. It is merged there and will land here if/when we get to backporting their customizations.

PR #562, opened 2022-01-23
Fix Series Category Affiliation

### LINKED ISSUE #551: Affiliation of category "Belongs to series" does not work
Categories set to "Belongs to series" are not shown in videos of the same series.

lkiesow:
This patch fixes series categories not showing up on other events of the
same series due to a parameter mismatch between front-end and back-end.

This fixes #551


JulianKniephoff:
This, too, was only buggy in the SWITCH version of the code. The community version actually consistently uses `series-extid`.

PR #565, opened 2022-01-24
Fix Scale Deletion

### LINKED ISSUE #541: When canceling the deleting of a scale the modal "Edit scale" gets closed
If I hit on "Delete" when editing a scale and then hit on "Cancel" the modal gets closed and I am back in the annotation tool. Specified was: The delete modal gets closed and I am back in the modal "Edit scale".

lkiesow:
This patch fixes the problem that no delete option was presented to
users in the scale editing dialog.

This fixes #541

### DIFF
--- a/frontend/js/views/scale-modal.js
+++ b/frontend/js/views/scale-modal.js
@@ -116,7 +116,9 @@ define([
                         ) })
                         .value()
                 );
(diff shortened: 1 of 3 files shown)


JulianKniephoff:
Merged in 27e1a2f738e373bf8a3db1ba81d95276dd7d0e61

PR #567, opened 2022-01-26
Improve the access and visibility descriptions

### LINKED ISSUE #533: Sharing option: Additional explanation for new option

lkiesow:
This patch updates the descriptions for the different access and
visibility options, making it more clear what the difference between
these options are.

Screenshot from 2022-01-27 00-27-15

This fixes #533


JulianKniephoff:
Merged in 675db24205a2298404590d20c13b06bb29ab3afa

PR #568, opened 2022-01-27
Fix deleting scales

lkiesow:
This patch fixes the problem that deleting scales will occasional fail
due to timing issues since the values belonging to the scale being
deleted are not yet deleted.

Instead of trying to get the timing right and waiting for requests in
the frontend, this patch fixes the problem in the backend, automatically
deleting all its values along with a scale.


JulianKniephoff:
Merged in b87aca26ddfd85a05fe8a6295dd6878320b921e0

PR #575, opened 2022-02-28
Make the tool work with Opencast 11

JulianKniephoff:
Note that the tests now no longer work with Opencast 10,
so if you want to build against that, you need to use something like

    mvn clean install ... -Dmaven.test.skip


lkiesow (approved):
Successfully installed the tool in Opencast 11.x :+1:

PR #576, opened 2022-03-04
`npm update --save`

JulianKniephoff:
Note that this, once more, might require something akin to

```
rm -rf frontend/node*
```


ebbertd:
Scheint zu funktionieren.

PR #589, opened 2022-10-11
Use composite videos in player if they exist

KatrinIhler:
This is achieved by preferring videos with composite/* flavor during sort.


JulianKniephoff (approved):
LGTM! :smiley:

PR #604, opened 2023-04-03
(2/3) Bugfixes

unterrichtsvideos:
This branch builds upon the [merged version](https://github.com/unterrichtsvideos/annotation-tool/tree/cc/master-merge) from PR #603 

The Bern (B, current master) and Münster (MS, questionnaire branch) versions had serveral bugs which were fixed after the merge, e.g.

- The export function was broken due to introduction of MCA (MS)

- A scale could not be edited (B)
image

- A shared category ("with all users") could not be edited from the same user (B)
image

- An assigned scale could not be used (B)
image

- The new edit menu was out of bounds / cut off (B)
image

see commits for further bug fixes


ChristianOellers:
⭐ Start reading here: Jul 25, 2022 - Commits after master merge #603

Arnei:
Played around with this a bit, here are  few things I've noticed:

- When creating an annotation from a questionnaire and setting the start or end time, a user can enter times that don't conform to the format `0:00:00`. The user is notified of this with a format error. But even then, you can still hit save and the Annotation Tool will make a best effort guess at what you were going for.
	- E.g. entering "60" is interpreted as "0:00:06". Not what I would have expected
- If a questionnaire has a public category and a private category as a label, it is treated as private. Is this expected behaviour?
- When switching to the questionnaire layout, the annotation list is not shown, even though the view dropdown claims that it is visible somewhere
  - Bildschirmfoto vom 2023-07-26 10-34-33
- Currently the only way to get the questionnaire view to show up is by selecting the questionnaire layout. It would be nice if I could add the view to my current layout just like e.g. "Loop controller".
- Unrelated to the questionnaire, categories (and their labels) that belong to a series may sometimes get duplicated. Not quite sure when this is happening, seems to be related to adding/removing videos from series.
- Bildschirmfoto vom 2023-07-26 10-28-39

unterrichtsvideos:
> * When creating an annotation from a questionnaire and setting the start or end time, a user can enter times that don't conform to the format `0:00:00`. The user is notified of this with a format error. But even then, you can still hit save and the Annotation Tool will make a best effort guess at what you were going for.  
>   * E.g. entering "60" is interpreted as "0:00:06". Not what I would have expected

Bug or feature? 🧐

Notes:
- In this context "Save" means "setting the annotation end". Trying to save the whole template-based annotation with an invalid value as annotation end will result in a format error note, asking to correct the value.
- Suggest: The behaviour of the annotation save button should be adopted for the "annotation end" - save button; this could be done after a merge 









> * If a questionnaire has a public category and a private category as a label, it is treated as private. Is this expected behaviour?
> * When switching to the questionnaire layout, the annotation list is not shown, even though the view dropdown claims that it is visible somewhere
>   
>   * Bildschirmfoto vom 2023-07-26 10-34-33
> * Currently the only way to get the questionnaire view to show up is by selecting the questionnaire layout. It would be nice if I could add the view to my current layout just like e.g. "Loop controller".
> * Unrelated to the questionnaire, categories (and their labels) that belong to a series may sometimes get duplicated. Not quite sure when this is happening, seems to be related to adding/removing videos from series.
> * Bildschirmfoto vom 2023-07-26 10-28-39

Arnei:
> Bug or feature? monocle_face
> 
> Notes for readers:
>
>    In this context "Save" means "setting the annotation end" (blue arrow, see image below). Trying to save the whole template-based annotation with an invalid value as annotation end will instead result in a format error note asking to correct the value.
    Suggest: The behavior of the annotation save button should be adopted for the "annotation end" - save button; this could be done after a merge as the other behavior could be tolerated for now?

Oh, so the save button for the whole template does do time stamp validation, did not know that. Arguably that is not very helpful, as it asks you to ignore the prominent "annotation end" save button. Which you cannot do if you also want to edit "annotation start". 

I'd be in favor of your suggestion, or just doing away with the "annotation end" save and cancel buttons

> Which branch did you review? The annotation list should be shown in [3/3 branch](https://github.com/opencast/annotation-tool/pull/605/commits); commit https://github.com/opencast/annotation-tool/commit/d089bea70623a43fe93be6d267f62d9e5863a9cc should have fix that (show the annotation list in annotation template view).

Reviewed this branch (2/3), thanks for telling me about the fix in 3/3.

> This should be a known bug originating from the current master version which introduced sharing categories with series leading to this duplication. As this bug should already exist in the master version and was not introduced by the merge, this should not be handled as a blocker for the merge but should be looked into later on.

Agreed, not a blocker then. Should have checked if that issues already existed, my bad.

unterrichtsvideos:
> I'd be in favor of your suggestion, or just doing away with the "annotation end" save and cancel buttons

+1 
I think it's not necessary to save the annotation start or end separatly. The implementation is based on the [original MCA concepts](https://github.com/opencast/annotation-tool/labels/mca) and code from the [old questionnaire branch](https://github.com/opencast/annotation-tool/tree/questionnaire).

> Agreed, not a blocker then. Should have checked if that issues already existed, my bad.

No worries, I'd suggest to check the described behavior in the master version again to make sure it's the same bug.

Issue #371, opened 2020-04-29
New annotation content type: Questionnaires 


ebbertd:
Currently the possible types of content for annotations are free text and structured annotations. Additionally instructors (anyone with admin rights in the annotation tool) should be able to create questionnaires. These questionnaires should consist of multiple questions followed by a free text. Students should only be able to use the content type questionnaire when a questionnaire has been defined by an instructor.

dagraf:
If I get it right, with adding questionnaires the annotation tool would change its main goal from analyzing video material to enriching educational videos with self-tests and quizzes. And I do not know, if we should go this way.
Anyhow, I think there is some thorough specification needed. For example (to only mention a few):

- What hapens with answers of students to these questions? Do they get saved as normal annotations but only "shared with administrator"?
- Should the teacher be able to indicate correct/wrong answers?
- Does the student receive a feedback how he performed? If yes, when?

ebbertd:
Thank you for your feedback. The questionnaires are supposed to be a help when analyzing that can be answered multiple times within a video. As the questionnaire would be a type of content.
As for your questions:
- The students answers should be saved as normal annotations. Therefore the sharing options are tied to the track.
- Indicating correct / wrong answers and giving feedback is not currently planned. Although, while not ideal, it could happen in the comments to the annotation.

dagraf:
Ok. But still a lot of questions remain and I'd like to repeat my suggestion to write a thorough specification. For example: How does an admin set up a questionnaire? How does a student answer these questionnaires? Is there an additional button? Or does an admin insert a questionnaire at a specific point in time and when reached, does the video stop and a modal gets opened showing the questions? ...

ebbertd:
We talked about this issue and the term `questionnaires` was quite misleading. The goal is not to provide a questionnaire in which students have to answers questions which are then marked as right or wrong. The goal is to enable instructors to create forms to guide their students through a previously specified annotation process. For this purpose instructors shall have the option to define an annotating form that can contain multiple instruction and annotation combination.

Issue #206, opened 2019-03-29
Include the annotation tools into iframe


Lyliian:
Hello,

I am trying to include the annotations tools page on Moodle by an iframe but the annotation tool returned an error and stay at "start loading video" : 

TypeError: Argument 1 of Window.getComputedStyle is not an object.

the annotation is working well without iframe.

Someone know how to solve this ? 

Regards,

Lylian.

JulianKniephoff:
Can you get a stack trace for that message/exception?

It might also be useful to see the code for how you are trying to embed the tool.

Lyliian:
Hi,

This is the iframe : 

<iframe allowfullscreen="" src="https://opencast.dsi.uca.fr/annotation-tool/index.html?id=ef554017-2a36-4493-a7da-fede9aa147a2" marginheight="0px" marginwidth="0px" width="1280" height="720" frameborder="0"></iframe>

And there is the trace of that message : 
![image](https://user-images.githubusercontent.com/12525724/55233387-b5932680-5228-11e9-95e6-766c527223ca.png)

Regards,

JulianKniephoff:
@Lyliian can you try whether #208 works for you? It fixed it on my end, at least.

Lyliian:
@JulianKniephoff Woooow ! It's working great.

Thank's a lot !

Best regards,

Lylian.

Lyliian:
@JulianKniephoff It's me again ! We got a problem with the iframe of annotation tool. After put the video in fullscreen and reduce it, the video size changed.

Before : 
![image](https://user-images.githubusercontent.com/12525724/55400436-2a7a9f00-554e-11e9-8cf5-e92246b6915f.png)

After : 
![image](https://user-images.githubusercontent.com/12525724/55400508-61e94b80-554e-11e9-989f-927cc38620ba.png)

It's not appaear when you press escape but just when you press the button for refuce the fullscreen.

JulianKniephoff:
And this happens only when you embed the tool?

Lyliian:
Absolutely !

JulianKniephoff:
I will look into it further when I have some time again. Can you do me a favor and create a new issue for this, and then reference it here?

PR #605, opened 2023-04-03
(3/3) Annotation template feature

unterrichtsvideos:
This branch builds upon the [merged+ bugfixed branch](https://github.com/unterrichtsvideos/annotation-tool/tree/cc/master-merge-bugfixes---REBASED) from PR #603 and PR #604 

:warning: This is the tested final version. However, if the bugfixes branch changes during QA and before it is merged into the final master, then this needs to be updated to.


frontend/js/access.js
```diff
@@ -37,6 +37,7 @@ define([
         PUBLIC: 1,
         SHARED_WITH_ADMIN: 2,
         SHARED_WITH_EVERYONE: 3,
+        SHARED_WITH_GROUP: 4,
```
unterrichtsvideos, line 40: This access level can be removed (this belonged to a draft course context feature which is not included in the merge)

frontend/js/collections/labels.js
```diff
@@ -58,9 +58,23 @@ define(
              */
             parse: function (resp) {
                 if (resp.labels && _.isArray(resp.labels)) {
-                    return resp.labels;
+                    // sort by abbreviation of label
+                    return resp.labels.sort((a, b) => {
```
unterrichtsvideos, line 62: sorting could also be simplified with a method to remove redundant code (see categories.js)

frontend/js/integrations/search.js
```diff
@@ -120,9 +127,11 @@ define([
             $.when(user, isAdmin).then(function (userResult, isAdmin) {
                 var user = userResult[0];
                 var userData = user.user;
+                var nickname = (typeof(userData.name) !== "undefined" && userData.name !== null) ? userData.name : userData.username;
+
                 this.user = new User({
                     user_extid: userData.username,
-                    nickname: userData.name || userData.username,
```
unterrichtsvideos, line 125: In our internal questionnaire branch, getting the real nickname was introduced earlier than in the master; this could be done either way i guess, the merge seems to remove masters approach which is a bit cleaner and could persist

frontend/js/integrations/search.js
```diff
@@ -174,7 +183,9 @@ define([
                     videos.map(function (track) {
                         return {
                             src: track.url,
-                            type: track.mimetype
+                            type: track.mimetype,
+                            framerate: track.video.framerate,
```
unterrichtsvideos, line 187: This also belongs to an extension, which is not part of this merge but seems to have slipped in, though I would suggest to leave it as it is, it may be useful anyway (we needed to hand over some more details for an extension)

frontend/js/models/annotation.js
```diff
@@ -46,15 +45,44 @@ define(
             defaults: function () {
                 return {
                     start: 0,
-                    duration: 0,
-                    comments: new Comments([], { annotation: this })
+
+                    // Value > 0 for time based annotation by default (contrary to point annotation)
+                    duration: 0.5,
```
unterrichtsvideos, line 50: As a fix could not be introduced in time into the library, this is a workaround fixing the usability issue while still preserving the feature bern wanted (-> create/convert annotations to point annotations should remain/exist)

Point annotations can be created by setting the same start and end time (copy &paste). as this would not be user friendly, the button was introduced to handle it for the user, defaults to non-point annotation, button converts to point annotation:
image

frontend/js/models/annotation.js
```diff
@@ -46,15 +45,44 @@ define(
...
             },
 
+            /**
+             * Constructor
+             * @alias module:models-annotation.Annotation#initialize
+             * @param {object} attr Object literal containing the model initialion attributes.
+             */
+            initialize: function (attr) {
+                // Fixes comments not being loaded/updated anymore (e.g. reload app); also fixes icon status.
+                // This was removed in master (35e244e000f82ff8def250bfe566bd213d4705a2) - Check if this is correct.
```
unterrichtsvideos, line 64: In master in questionnaire the comment loading behaviour seems a bit weird at this point, this should be checked in the review I guess

frontend/js/models/track.js
```diff
@@ -46,7 +46,7 @@ define([
              * @static
              */
             defaults: {
-                access: ACCESS.PRIVATE,
+                access: ACCESS.SHARED_WITH_ADMIN,
```
unterrichtsvideos, line 49: We discussed that in the public meetings as we had usability issues in classrooms (in courses, the students tend to forget to share it with admins, so the work could frequently not be reviewed)

frontend/js/player-adapter-html5.js
```diff
@@ -173,6 +173,11 @@ define(
                             e.preventDefault();
                         });
 
+                        // Workaround to fix chrome fullscreen bug (#214)
+                        $(document).bind('fullscreenchange webkitfullscreenchange mozfullscreenchange msfullscreenchange', function (e) {
+                            self.resetSize();
```
unterrichtsvideos, line 178: The video does not scale/resize correctly in browsers; see also https://github.com/opencast/annotation-tool/issues/206#issuecomment-478961780

This reduces the issue

frontend/js/views/main.js
```diff
@@ -494,8 +577,10 @@ define(
                 goldenLayout.on("stateChanged", function (event) {
                     localStorage.setItem("layout-custom", JSON.stringify(goldenLayout.toConfig()));
                     localStorage.setItem("layout", "custom");
+                    annotationTool.playerAdapter.resetSize();
                 });
                 goldenLayout.init();
+                this.goldenlayout = goldenLayout; // TODO do not expose?
```
unterrichtsvideos, line 583: Makes goldenLayout Object accessible from annotationTool object to use several api functions to improve usability; should be fine to add it

frontend/js/views/timeline.js
```diff
@@ -73,22 +73,38 @@ define([
...
+            .value()
+            .join(", ");
+
+        var color = annotation.getColor();
+        item.style = color && (
+            "background-color:" + color + ";" +
                 "color:" + (
                     chroma(color).luminance() < 0.5
-                        ? "white"
+                        ? "black" // before: white
```
unterrichtsvideos, line 90: Colors wouldn't work otherwise with MCA

opencast-backend/annotation-impl/src/main/java/org/opencastproject/util/osgi/SimpleServicePublisher.java
unterrichtsvideos, line 1: Obsolete dependency

unterrichtsvideos:
> * > This branch builds upon the [merged+ bugfixed branch](https://github.com/unterrichtsvideos/annotation-tool/tree/cc/master-merge-bugfixes---REBASED) from PR [(1/3) Merge questionnaire into master #603](https://github.com/opencast/annotation-tool/pull/603) and PR [(2/3) Bugfixes #604](https://github.com/opencast/annotation-tool/pull/604)
>   
>   
>   that does not seem to be true? This includes [(1/3) Merge questionnaire into master #603](https://github.com/opencast/annotation-tool/pull/603) alright, but it diverges from [(2/3) Bugfixes #604](https://github.com/opencast/annotation-tool/pull/604).

Comparing the commit history, the 3/3 branch (unterrichtsvideos:cc/master-merge-bugfixes-annotationsvorlagen---SQUASHED) was created from branch unterrichtsvideos:cc/master-merge-bugfixes---REBASED and the commit [Fix missing translations, consistent language files (a6e0319)](https://github.com/opencast/annotation-tool/pull/604/commits), 4 non-conflicting commits were later added to the bugfixes branch. Otherwise the following 4 commits have to be added / merged to the 3/3 branch as well.

- Disallow VSCode history (plugin, native) and browser DevTools files
- Cleanup comments, remove debug logs
- Exclude Git patch files
- Fix scrolling behaviour of category list

> * Some of the comments already made by @unterrichtsvideos seem to suggest changes. Will someone make those, still?

These comments were part of a quick review and parts I stumbled upon. If that's fine for you, feel free to make these changes if you agree. Feel free to make any changes in the code regarding to the goal to create the merged version. 

> * Can you add an example questionnaire JSON thingy (along with the necessary categories) here for testing? Or multiple if you have them? I remember having access to some folder, but I can't seem to find it right now. 🙈

the [public meeting](https://pad.uni-muenster.de/2rRnVcAJRl-ezIcdVeslZA#2023-02-07-3-pm-CET) contains a link to [test files](https://uni-muenster.sciebo.de/s/vbmsXj9sKQFSGtY/download?path=%2FCommunity%2FTestdaten&files=testdaten.zip), don't hesitate to contact us, whether these are working (they should) or need and update.

frontend/js/views/timeline.js
Arnei: Does MCA stand for "Multi Colored Annotation"? Is that an annotation with multiple labels of differing colors?

This code makes the conditional redudant, as the color will now always be black. Which coincidentally works, as none of the predefined category colors are dark enough to make black unreadable I suppose (at least that's what I think this is about?). Relying on the predefined category colors not changing seems suboptimal though.

unterrichtsvideos: [Multi-Content-Annotation](https://github.com/opencast/annotation-tool/labels/mca)
(previously: annotation was either "free text" or "label", now: annotation can contain multiple "free text" and/or "labels")

I remember there was a reason with this change regarding colors of MCA and display style of annotations in timeline;
I think the conditional was kept with the comment as it was unsure whether this would require another change later on
we could figure out what value(s) work in all (or most) cases and adjust it again

Arnei: Thanks for the explanation. I looked into it a bit, and unfortunately there seems to be no quick fix for this, at least I am unable to come up with one. Since this code works with the current state of the annotation tool, albeit coincidentally, I think we can leave this as is for now (I did add a todo to the code though).

Arnei (commented):
Is there documentation on how to use questionnaires? Might be me being blind, but I can't find any in this PR. And since users must use a template to use questionnaires, this feature is unusable if you don't know what a template must look like.

unterrichtsvideos:
> Is there documentation on how to use questionnaires? Might be me being blind, but I can't find any in this PR. And since users must use a template to use questionnaires, this feature is unusable if you don't know what a template must look like.

Indeed we owe a documentation for this. We will create and provide one when the merged version is finalized.

Following the discussion regarding the term "[questionnaire](https://github.com/opencast/annotation-tool/issues/371#issuecomment-631335957)", for end users the feature is called "annotation templates", as a questionnaire is basically a "annotation-template"-based form editor for creating MCA with a pre-defined structure (which can be used e.g. to create annotation tasks for students); internally the term was not refactored

It was planned that templates can be created with a "questionnaire" builder (template builder)

Related project (unfinished state): Questionnaire builder
Code: https://github.com/luniki/oat-questionnaire-creator
Test-Deploy: https://www.uni-muenster.de/Unterrichtsvideos/builder/

Arnei (approved):
I'd say this is good enough to merge. Still needs to documentation, and did not go through the entire code (it's simply too much), but as far as I can tell the annotation templates work, don't break anything and this PR even fixes some previously broken things! 

We'll still want to wait on @JulianKniephoff 's opinion before merging though.

PR #610, opened 2023-07-13
Just `stylelint` things

JulianKniephoff:
Updates `stylelint` in two steps:

* To `^14` adding the necessary dependencies and configuraiton to make it understand `*.less` files (again).
* To `^15` migrating the (now deprecated) stylistic rules to a plugin

Supersedes #608.


Arnei (approved):
Seems to work fine. Although arguably not very useful, as it does not show up in my IDE (VS Code) and is not terribly prominent in the build logs either, but that's not the fault of this PR.

PR #611, opened 2023-07-13
Update `h2`

JulianKniephoff:
See commit messages for "details."

Supersedes #609.

PR #614, opened 2023-07-28
Update `h2`

JulianKniephoff:
See #611.

Unfortunately this approach doesn't work for Postgres.

Re-requesting as draft for now to remind me to fix this eventually.


JulianKniephoff:
Note to self: Check whether dependabot will make automatic updates for this again once we update it manually. For now I told it to ignore `h2`: https://github.com/opencast/annotation-tool/pull/617#issuecomment-1854369845

JulianKniephoff:
Actually this shouldn't be a PR. 🤷‍♀️ See #619.

PR #622, opened 2024-01-11
Update migration script

oellers:
This script should work for different versions of the tool (bern-branch, questionnaire-branch) to migrate to the merged version "next".


JulianKniephoff (approved):
This looks good to me at a first glance. Haven't tested it, yet, but will (have to) do so when I migrate Bern. I reserve the right to adapt it while doing so, but until then/for that this is more useful to have in the repo instead of sitting around in a PR.

Thanks! :)

PR #623, opened 2024-03-14
Make tool work with Solr removal

Arnei:
Over in Opencast, there is a pull request that makes changes to the Opencast Search Service, breaking the Annotation Tool:
https://github.com/opencast/opencast/pull/5597

Since that pull request 5597 is all but assured to go into Opencast 16, this commit attempts to fix
the errors that 5597 would introduce.

This DOES constitute a breaking change.

### DIFF
--- a/opencast-backend/annotation-impl/src/test/java/org/opencast/annotation/endpoint/TestRestService.java
+++ b/opencast-backend/annotation-impl/src/test/java/org/opencast/annotation/endpoint/TestRestService.java
@@ -22,9 +22,6 @@ import org.opencast.annotation.api.ExtendedAnnotationService;
 import org.opencast.annotation.impl.persistence.ExtendedAnnotationServiceJpaImpl;
 
 import org.opencastproject.mediapackage.MediaPackage;
-import org.opencastproject.search.api.SearchQuery;
-import org.opencastproject.search.api.SearchResult;
-import org.opencastproject.search.api.SearchResultItem;
 import org.opencastproject.search.api.SearchService;
 import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.DefaultOrganization;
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -67,8 +67,8 @@
 
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-    <opencast.build.version>14.0</opencast.build.version>
-    <opencast.deploy.version>;version="${opencast.build.version}"</opencast.deploy.version>
+    <opencast.build.version>16-SNAPSHOT</opencast.build.version>
+    <opencast.deploy.version>;version=16.0</opencast.deploy.version>
```
JulianKniephoff, line 71: Note to self: This is going to have to change before this goes in, once the Opencast PR is merged and released.

opencast-backend/annotation-impl/src/test/java/org/opencast/annotation/endpoint/TestRestService.java
```diff
@@ -90,17 +87,12 @@ private static AuthorizationService getAuthorizationService() {
...
-    SearchResult searchResult = EasyMock.createNiceMock(SearchResult.class);
-    EasyMock.expect(searchResult.getItems()).andReturn(new SearchResultItem[]{searchResultItem}).anyTimes();
-    EasyMock.replay(searchResult);
-
     SearchService searchService = EasyMock.createNiceMock(SearchService.class);
-    EasyMock.expect(searchService.getByQuery(EasyMock.anyObject(SearchQuery.class)))
-            .andReturn(searchResult).anyTimes();
+    try {
+      EasyMock.expect(searchService.get(EasyMock.anyObject(String.class))).andReturn(mediaPackage).anyTimes();
+    } catch (Exception e) {
```
JulianKniephoff, line 93: I would sleep ever so slightly better if we just caught whatever we have to so we don't hide any real exceptions.

JulianKniephoff (changes requested):
Code looks good apart from minor nitpicks. Compiled and tested (very rudimentarily) as well, which worked fine.

JulianKniephoff (approved):
Fixed the nitpicks myself, and tested it locally.

We are waiting for the 16.0 jars to be in central; without those, CI won't be able to build the tool.

JulianKniephoff:
Jars are there now, let's gooo!

PR #627, opened 2024-05-16
Fixes for serverless HLS

KatrinIhler:
Actual changes from @luniki.

- Pefer master playlists over non-master playlists and mp4s
- Prefer native_hls renderer
- Wait for !isFinitite instead of isNan for this renderer

### DIFF
--- a/frontend/js/integrations/search.js
+++ b/frontend/js/integrations/search.js
@@ -159,6 +159,7 @@ define([
                         RegExp.prototype.test.bind(/application\/.*|video\/.*/),
                         _.property("mimetype")
                     ));
+                videos = _.sortBy(videos, "master").reverse();
                 videos.sort(
                     util.lexicographic([
                         util.firstWith(_.compose(
--- a/frontend/js/player-adapter-html5.js
+++ b/frontend/js/player-adapter-html5.js
@@ -85,7 +85,7 @@ define(
 
                 window.Hls = Hls;
                 mediaElementPlayer = new mejs.MediaElementPlayer(targetElement, {
-                    renderers: ["html5", "native_hls"],
+                    renderers: ["native_hls", "html5"],
                     alwaysShowControls: true,
                     autoRewind: false,
                     stretching: "fill",
@@ -96,7 +96,7 @@ define(
                          */
                         $(mediaElement).on("canplay durationchange", function () {
                             // If duration is still not valid
-                            if (isNaN(self.getDuration()) || mediaElement.readyState < 1) {
+                            if (!isFinite(self.getDuration()) || mediaElement.readyState < 1) {
                                 return;
                             }
 


frontend/js/integrations/search.js
```diff
@@ -159,6 +159,7 @@ define([
                         RegExp.prototype.test.bind(/application\/.*|video\/.*/),
                         _.property("mimetype")
                     ));
+                videos = _.sortBy(videos, "master").reverse();
```
JulianKniephoff, line 162: Can we integrate this in the `sort` below? Something like

```js
videos.sort(
    util.lexicographical([
        util.firstWith(
            _.property("master")
        ),
        // [...]
    ])
);
```

Should work, but please double-check me on this.

frontend/js/player-adapter-html5.js
```diff
@@ -85,7 +85,7 @@ define(
 
                 window.Hls = Hls;
                 mediaElementPlayer = new mejs.MediaElementPlayer(targetElement, {
-                    renderers: ["html5", "native_hls"],
+                    renderers: ["native_hls", "html5"],
```
JulianKniephoff, line 88: Is this necessary even with the correct sort order of the videos? 🤔

Looking at the docs for this setting I also wonder: Can we get away with not specifying this at all?

frontend/js/player-adapter-html5.js
```diff
@@ -96,7 +96,7 @@ define(
                          */
                         $(mediaElement).on("canplay durationchange", function () {
                             // If duration is still not valid
-                            if (isNaN(self.getDuration()) || mediaElement.readyState < 1) {
+                            if (!isFinite(self.getDuration()) || mediaElement.readyState < 1) {
```
JulianKniephoff, line 99: Wouldn't live streams have infinite durations? 🤔 Although to be fair, the rest of the tool can't deal with those, either. x)

Still, I wonder: Did you encounter a case where the duration is infinite first, and then changes to something finite?

JulianKniephoff (changes requested):

JulianKniephoff:
Pinging @luniki to look at this review since he is the original author of the patch IIRC.

frontend/js/player-adapter-html5.js
marwyg: I removed the `renderers` line. I read the documentation and this field doesn't seem mandatory. It still works when I removed it: https://github.com/opencast/annotation-tool/pull/646

frontend/js/player-adapter-html5.js
marwyg: As mentioned in https://github.com/opencast/annotation-tool/pull/646, I didn't see such a case, but Marcus did, I don't know. But I compared in my other PR both functions (isNaN and !isFinite) and I think it should be okay like this.

marwyg:
I took the changes from here, addressed @JulianKniephoff review points and made my own PR https://github.com/opencast/annotation-tool/pull/646. Maybe someone can take a look again, I think this should be okay now.

JulianKniephoff:
Closing in favor of #646

PR #629, opened 2024-09-09
fix loading when 'video' key is missing

jduehring:
After updating to OC 16.4 we had the issue that the videos did not load in the OAT. @luniki found out that it is related to HLS tracks missing the "video" key. The changes in this PR fix this by only using tracks that have the correct keys. 

We are however unsure as to why this was no issue with older versions of the OAT.

### DIFF
--- a/frontend/js/integrations/search.js
+++ b/frontend/js/integrations/search.js
@@ -184,8 +184,8 @@ define([
                         return {
                             src: track.url,
                             type: track.mimetype,
-                            framerate: track.video.framerate,
-                            resolution: track.video.resolution
+                            framerate: track.video && track.video.framerate,
+                            resolution: track.video && track.video.resolution
                         };
                     })
                 );


JulianKniephoff:
> We are however unsure as to why this was no issue with older versions of the OAT.

This is probably because accessing/passing along these properties is new in the `next` branch: https://github.com/opencast/annotation-tool/commit/8397285fc1a68e04900c4e1dc795596be3c22d4a

I don't know why it was added in the first place, maybe @ComeIn-NRW @github-canni can shed some light?

JulianKniephoff:
I took the liberty of actually removing the fields since I could find no good reason for us to pass them around in MediaElement.js. 🤷‍♀️ They probably came in as part of some unfinished business during the "Münster-merge."

cf. bf39a876494ce21a9b6db7094e420bc199d72607

oellers:
I can shed some light on this.
The commit comes from an internal branch that made the annotation tool xAPI-compatible.
This required frame rate and resolution (as we built upon the [xapi cop video profile](https://liveaspankaj.gitbooks.io/xapi-video-profile/content/statement_data_model.html)).
At some point this was functional as we tested the xAPI statements.

If there's interest in xAPI for the annotation tool, we might work on an implementation for the next version of the tool at some point in the future.

JulianKniephoff:
Interesting, thanks for the info! :D

Issue #633, opened 2024-11-26
Title of video does not get shown


dagraf:
Branch: Stable

At the very top of the annotation tool the title of the opened video should be shown. At the moment it always says "Video ohne Titel".

jduehring:
@JulianKniephoff You can assign me here!

PR #637, opened 2024-11-28
Fix wrong dcTitle key

jduehring:
This PR fixes https://github.com/opencast/annotation-tool/issues/633 . 

The structure of the field `result.dcTitle` changed at some point in OC, so that we need to call the sub-field `dc.title` .

### DIFF
--- a/frontend/js/integrations/search.js
+++ b/frontend/js/integrations/search.js
@@ -115,7 +115,7 @@ define([
                 return {
                     video_extid: id,
                     series_extid: mediaPackage.series,
-                    title: result.dcTitle
+                    title: result.dc.title
                 };
             });
         },
--- a/frontend/js/integrations/search.js
+++ b/frontend/js/integrations/search.js
@@ -111,11 +111,11 @@ define([
          * @return {Promise.<object>} Metadata about the video
          */
         getVideoParameters: function () {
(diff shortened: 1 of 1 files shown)


dagraf:

JulianKniephoff (changes requested):
This works, but looking at this part of the code more closely again, we could actually replace `getVideoParameters` with the following, which is a bit simpler **and** should be portable across the breaking changes to the search API:

```js
        getVideoParameters: function () {
            return mediaPackage.then(function (mediaPackage) {
                return {
                    video_extid: mediaPackage.id,
                    series_extid: mediaPackage.series,
                    title: mediaPackage.title,
                };
            });
        },
```

jduehring:
> This works, but looking at this part of the code more closely again, we could actually replace `getVideoParameters` with the following, which is a bit simpler **and** should be portable across the breaking changes to the search API:
> 
> ```js
>         getVideoParameters: function () {
>             return mediaPackage.then(function (mediaPackage) {
>                 return {
>                     video_extid: mediaPackage.id,
>                     series_extid: mediaPackage.series,
>                     title: mediaPackage.title,
>                 };
>             });
>         },
> ```

Agreed. I had this in mind as well but thought there was probably some deeper reason for getting the info from different fields, which I am overlooking :D Anyway, changed the code and it seems to work as expected.

JulianKniephoff (approved):

Issue #630, opened 2024-11-21
Options to stop video when writing/setting annotations are hidden


dagraf:
Branch: stable

If I open a video in the annotation tool the options to stop the video are hidden behind the list of annotations because they are hidden behind the lists of annotations. When I enlarge the tab/window "Create new annotation", the options is shown. When I enlarge it even more, the 'white box' reserved for categories gets higher. When making the tab smaller, the white box stays as big as it was.

Expected solution (see screencast: wished behavior on the right):
Depending on the size of the tab/window "Create new annotation" the 'white box' should shrink/enlarge so that the options to stop the video when writing/setting annotations is always visible.

jduehring:
@JulianKniephoff You can assign me here as well :D

dagraf:
It now works as expected. Thx for the fix!

PR #638, opened 2024-12-02
Fix height of container for creating new annotations

jduehring:
This PR should fix https://github.com/opencast/annotation-tool/issues/630 . 

For some reason the default height for the container containing the labels was set to `100vh`. I set this to `auto` so that it won't use the whole screen height. 

I also changed the value for `overflow-y` to `auto` so that the scrollbar for labels is only visible, when neccessary.

### DIFF
--- a/frontend/style/annotations/annotate.less
+++ b/frontend/style/annotations/annotate.less
@@ -30,8 +30,8 @@ i.check {
         flex-direction: column;
 
         .categories-container-scroll {
-            overflow-y: scroll;
-            height: 100vh;
+            overflow-y: auto;
+            height: auto;
         }
 
         .dropdown {


dagraf:

JulianKniephoff (changes requested):
The culprit here was 61478c92a6bfe0c7d973bbb8a3747c1f321548f1.

However, I see no difference between what this PR does, and not applying this class at all, in which case we could get rid of it, too, since it isn't used anywhere else.

However **again** I'm not sure the result is what we want: The problem is (still/again) that when you have many labels in one category, the white container expands to accommodate them, potentially pushing the checkboxes outside the visible area and making you scroll, instead of having to scroll within the white container itself to reach all the categories.

@dagraf can you clarify what the requirements are?

dagraf:
@JulianKniephoff The requirement as I see it would be: "A user can see the options to stop the video when this user does freshly load a video."
I can live with the solution as it is right now with but I am open to improvements. Do you have here an improvement in mind that could get implemented easily?

dagraf:
Right. But as you do not have a specific improvement in mind which can meet the stated requirement and as I said that I can live with the current solution: We implement this solution and close this ticket.

JulianKniephoff:
Fair. :)

I would still say that we should just get rid of `.categories-container-scroll` in that case. Except @jduehring tells me it **does** actually change things.

jduehring:
@JulianKniephoff Agreed, I removed the container and as you said it doesn't seem to change anything else.

PR #639, opened 2024-12-02
Put Less into math mode `'always'`

### LINKED ISSUE #632: Adding content and editing an existing annotation (with the pen icon) is not possible
Branch: Stable

When I have an existing annotation and have it expanded: There is a "Pen"-Icon and a button "Inhalt hinzufügen". When I click on the pen or the button, nothing happens and the tool can not be used anymore. All clicks have no effect and a browser reload is needed.

I assume this has something to do with the new feature "Annotationsvorlagen".

JulianKniephoff:
The upgrade to v4 of Less in b154c7b8a6767303fd9e9a443373d3c8f13faa33 introduced some changes that broke our version of Bootstrap. Specifically, division in `.less` files now has to be in parantheses by default. We fixed a few instances of that in the vendored files, but missed many others.

Instead of futzing around with these files even more, this patch puts Less in a compatibility mode. **Note**: This mode is deprecated, so this is only a temporary solution, until we find the time and money to update Bootstrap.

Fixes #632.

### DIFF
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -17,6 +17,7 @@ module.exports = function (grunt) {
                 syncImport: true,
                 compress: true,
                 sourceMap: true,
+                math: 'always',
                 sourceMapURL: 'style/style.css.map',
                 sourceMapBasepath: 'style'
             },


jduehring:
Looks good to me. Tested editing existing annotations, adding new content as described in the issue as well as various other actions. I did not experience any freezes anymore.

dagraf:

Issue #634, opened 2024-11-26
Categories belongig to a series: Labels show up twice in most videos 


dagraf:
Branch: Stable

Precondition: Have a series with multiple videos, e.g. A, B and C

Steps to reproduce:
1. Create a category "Zur Serie" in video A with to labels (S1 and S2) and choose the option "Belongs to series".
2. Open video B. See that everything is fine. Meaning: The category is shown with its to labels.
3. Reload the browser OR close the video B and open it newly.
4. See error: All lables of this category are shown twice in video B (see screenshot). In video A (where the category got created) the labels are shown correctly. 
5. If I edit this category "Zur Serie " in video B all labels also show up twice (see screenshot). When I edit the same category in video A (where I created it) the labels are correctly shown (only once).

<img width="1061" alt="Bildschirmfoto 2024-11-26 um 09 50 19" src="https://github.com/user-attachments/assets/73f3aa29-276d-48bb-9457-59b224f26a74">

<img width="541" alt="Bildschirmfoto 2024-11-26 um 09 50 31" src="https://github.com/user-attachments/assets/c31fc48e-7411-4c0e-85b6-0b50511acb06">

dagraf:
See also this issue for a similar (or maybe even identical) problem: https://github.com/opencast/annotation-tool/issues/601

jduehring:
@JulianKniephoff You can assign me here as well!

Issue #601, opened 2023-02-20
Category affiliation: If a category affiliation gets changed from series to video in the 'other' event labels end up showing up mutliple times


dagraf:
Steps to reproduce:
1. Have a series with at least two videos (A and B).
2. In video A: Create a category with at leas one label and set its affiliation to "Belonging to series".
3. In video B: Edit this category's affiliation to "Belongig to series". Click on save.
4. Still in video B: See that the labels get shown multiple times.
5. Still in video B: After a reload of the browser the labels do not get shown multiple times. But when opening the "Edit category" window the labels still show up multiple times.

After step 2 (see category "S - Teacher1"):
<img width="629" alt="Bildschirm­foto 2023-02-20 um 21 57 20" src="https://user-images.githubusercontent.com/15246287/220197394-28d2496e-0218-45e9-a4fa-2e7c1ea7958a.png">

After step 4:
<img width="641" alt="Bildschirm­foto 2023-02-20 um 21 58 10" src="https://user-images.githubusercontent.com/15246287/220197492-07dc3c48-0979-4ba4-b9e1-ba9ba28ada33.png">
<img width="637" alt="Bildschirm­foto 2023-02-20 um 21 58 23" src="https://user-images.githubusercontent.com/15246287/220197524-7c2a7636-9414-4054-8c03-6052636eba9c.png">

After step 5:
<img width="644" alt="Bildschirm­foto 2023-02-20 um 21 58 41" src="https://user-images.githubusercontent.com/15246287/220197554-601ab85a-a94b-4bcb-81af-5f745ef933f7.png">
<img width="625" alt="Bildschirm­foto 2023-02-20 um 21 58 55" src="https://user-images.githubusercontent.com/15246287/220197575-ee7eb3a7-5144-4e57-959f-d0cfdbfa9dea.png">

JulianKniephoff:
So yes, this is very likely caused by the same bug as #645. This is a bigger problem, so I'll close this, and likely the other as well, and replace them with a new issue that captures the full extent of the problem.

PR #645, opened 2024-12-20
Categories belongig to a series: Duplicated Labels

jduehring:
This PR-Draft relates to https://github.com/opencast/annotation-tool/issues/634 and https://github.com/opencast/annotation-tool/issues/601 . 

I "fixed" the current implementation in a way that the labels view now should not be showing duplicated labels anymore.

However, the implemenation is still majorly flawed as indicated in the comments: 
https://github.com/jduehring/annotation-tool/blob/0cb527bdaf654e287a1f4ecbfa927d0f9a546bc3/opencast-backend/annotation-impl/src/main/java/org/opencast/annotation/impl/persistence/ExtendedAnnotationServiceJpaImpl.java#L967C9-L969C145

When updating the labels for a series from the master series category, they currently are beeing deleted and newly created each time the app gets refreshed. But since the app tries to keep track of deleted labels, they essentially get duplicated in the background (which can still be seen in the "edit category" view) .  

We could probably try to filter all deleted labels and only show the newly created ones, but that doesnt fix the underlying issue of creating more and more (deleted) labels each refresh. Ideally the labels should only be created once and only update, if new lables have been created in the master series category. Unfortunately that does'nt seem as straight forward as it sounds. 

I could not implement a working, satisfying solution yet and will discuss this matter with @JulianKniephoff and/or @mwygas in January to hopefully make quick progress here.

### DIFF
--- a/opencast-backend/annotation-impl/src/main/java/org/opencast/annotation/impl/persistence/ExtendedAnnotationServiceJpaImpl.java
+++ b/opencast-backend/annotation-impl/src/main/java/org/opencast/annotation/impl/persistence/ExtendedAnnotationServiceJpaImpl.java
@@ -916,6 +916,7 @@ public final class ExtendedAnnotationServiceJpaImpl implements ExtendedAnnotatio
           throws ExtendedAnnotationException {
 
     List<Label> labels;
+    List<Label> deletedLabels = new ArrayList<>();
 
     if (since.isSome()) {
       List<LabelDto> labelDtos = findAllWithOffsetAndLimit(LabelDto.class, "Label.findAllOfCategorySince", offset, limit,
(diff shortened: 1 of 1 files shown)


dagraf:
Thx for the update. I'm looking forward to hearing here about the results of your discussion.

JulianKniephoff:
@dagraf can you quickly confirm or refute something for me?

When we switch a series category to a video category, all the videos currently in the series are supposed to keep said category around, right?

dagraf:
> @dagraf can you quickly confirm or refute something for me?
> 
> When we switch a series category to a video category, all the videos currently in the series are supposed to keep said category around, right?

No, they should loose the said category. This is what happens right now on our productive annotation tool installation and I would expect to keep this behavior. If for example, an admin makes a category belonging to series by mistake and changes it later to an event category, this admin would be forced to delete the kept categories in all videos.

But: What we should keep are the already set annotations which where created using this said category.

What is your opinion here?

JulianKniephoff:
Interesting, but that's good, that would simplify things, thanks!

I'm still working my way through all the intricacies of the problem, this one's a real doozy. 😵‍💫

The changes required to fix this properly will be rather extensive, though, and while I try to work these out I'm also looking out for any kind of quick fix we can apply in the meantime.

I'll got back to y'all here once I have a grasp of the full picture.

PR #646, opened 2025-01-24
Fixes for serverless HLS

marwyg:
I made my own PR, I hope that's okay. The original draft is here: #627 
This PR contains the actual changes from @luniki + changes after review from @JulianKniephoff. The review is also to be found here: #627.

The 3 points Julian made:
1. Integrating the sort into the sorting below. I did that and it works. (https://github.com/opencast/annotation-tool/pull/627#discussion_r1745519948)
2. Is it possible to remove `renderers: ["html5", "native_hls"]`completly?: Yes, I removed it and it works. (https://github.com/opencast/annotation-tool/pull/627#discussion_r1745529603)
3. I didn't see a case where the duration was infinity, but that doesn't mean that Marcus didn't see something. I testet a small javascript function to see, how !isFinite and isNaN differs from each other:

```javascript
function compareIsNaNAndIsFinite() {
  const testValues = [NaN, "hello", undefined, "123", 123, Infinity, -Infinity, null, true, false, ""];

  console.log("Value\t\tisNaN(value)\t!isFinite(value)");
  console.log("------------------------------------------------");
  testValues.forEach(value => {
    console.log(
      `${String(value).padEnd(10)}\t${isNaN(value)}\t\t${!isFinite(value)}`
    );
  });
}

compareIsNaNAndIsFinite();
```
### The result:
```
Value		isNaN(value)	!isFinite(value)
------------------------------------------------
NaN       	true		true
hello     	true		true
undefined 	true		true
123       	false		false
123       	false		false
Infinity  	false		true
-Infinity 	false		true
null      	false		false
true      	false		false
false     	false		false
          	false		false
```
So I think, it's okay to use !isFinite() instead, because it can handle one more case, but behaves the same in the other cases.

I added another small line from Marcus, where he triggers an update for the video duration. It was reported, that the time wasn't displayed correctly.

### DIFF
--- a/frontend/js/integrations/search.js
+++ b/frontend/js/integrations/search.js
@@ -159,6 +159,7 @@ define([
                         RegExp.prototype.test.bind(/application\/.*|video\/.*/),
                         _.property("mimetype")
                     ));
+                videos = _.sortBy(videos, "master").reverse();
                 videos.sort(
                     util.lexicographic([
                         util.firstWith(_.compose(
--- a/frontend/js/integrations/search.js
+++ b/frontend/js/integrations/search.js
@@ -159,9 +159,11 @@ define([
                         RegExp.prototype.test.bind(/application\/.*|video\/.*/),
                         _.property("mimetype")
                     ));
-                videos = _.sortBy(videos, "master").reverse();
                 videos.sort(
                     util.lexicographic([
+                        util.firstWith(
+                            _.property("master")
+                        ),
                         util.firstWith(_.compose(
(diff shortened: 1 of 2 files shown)


marwyg:
I tested this with the master and the stable branch.

JulianKniephoff:
> I added another small line from Marcus, where he triggers an update for the video duration. It was reported, that the time wasn't displayed correctly

@marwyg do you have a reference for that? Like an issue or something? 🤔

PR #650, opened 2025-09-29
Filter deleted labels

### LINKED ISSUE #649: Filter deleted labels in the frontend
With the introduction of MCA, annotations no longer contain a copy of their label's data (title, color, ...). They merely refer to the label by its ID. That means we can no longer delete labels. Instead we just mark them as such.

The problem is that the frontend wasn't updated to know about this. Without a larger refactoring effort, it still needs to fetch all labels, even the deleted ones. And this is what happens, but it uses these labels in the structured annotation view and category editor as well (and maybe in other places like the MCA modals), and so these need to be filtered in all the appropriate places.

JulianKniephoff:
Will close #649.

### DIFF
--- a/frontend/js/views/category-modal.js
+++ b/frontend/js/views/category-modal.js
@@ -170,7 +170,12 @@ define([
             this.model = options.model;
 
             this.labels = new Labels(
-                this.model.get("labels").invoke("clone"),
+                this.model.get("labels").chain()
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
OK, I think I checked all relevant places now, and this was the only one where this was forgotten. Marking as ready.

